PR #1, opened 2026-01-19
Added Copy JSON Button

IIITM-Jay:
### Added Copy JSON button

Added `Copy JSON {}` button beside the existing `Copy` button to enable users to copy the JSON format of the instruction as well.

### Results

Screenshot from 2026-01-19 17-43-20

The copied json will look like
```
{
  "extension": "RV64I",
  "description": "Standard Integer Base (64-bit)",
  "use": "Servers, Mobile, PC",
  "reference": "https://github.com/riscv/riscv-isa-manual",
  "mnemonic": "SLLI",
  "encoding": "000000-----------001-----0010011",
  "variable_fields": [
    "rd",
    "rs1",
    "shamtd"
  ],
  "extension_tags": [
    "rv64_i"
  ],
  "match": "0x1013",
  "mask": "0xfc00707f"
}
```

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -7,6 +7,7 @@ import {
   ArrowRight,
   ArrowUpRight,
   Copy,
+  Braces,
   ChevronLeft,
   ChevronRight,
 } from 'lucide-react';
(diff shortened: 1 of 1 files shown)


IIITM-Jay:
@rpsene requesting your feedback on above. Thanks!

rpsene:
Thanks for this, and I am sorry it sat since January without a reply. That is on us, not on you.

Closing it because the export paths it adds are now covered. The builder has Export Configuration YAML, the Encoder Validator has Copy report, instruction details have a copy button, and the -march string has its own copy. YAML rather than JSON is deliberate: it is the format riscv-config consumes, so it is directly usable by a toolchain.

The branch also conflicts with main now. If there is a specific case where JSON would be more useful than the YAML export, say so and I will look again.

PR #11, opened 2026-04-23
Tooling: add extension coverage report command (mapped/unmapped and group-wise summary)

### LINKED ISSUE #10: Tooling: generate extension coverage report (mapped/unmapped and group-wise summary)
## Problem
There is no single command that reports mapping coverage and remaining gaps.
This makes progress tracking and milestone planning difficult.

## Proposed work
Add a report command/script that outputs:
- total extensions
- extensions with instruction maps vs without
- group-wise coverage (`base`, `z_*`, `s_*`)
- extension IDs present in catalog but missing instruction mapping
- unresolved opcode tags (if available from reconciliation step)

Mapping rule:
- mapped: `instructions` is an object with at least 1 key
- unmapped: `instructions` missing, not an object, or empty

## Acceptance criteria
- One command generates a machine-readable report (JSON).
- Optional human-readable report (Markdown/text) with summary table.
- Document command usage in README.

amkr6207:
## Summary
This PR adds tooling to generate extension coverage reports from the current catalog data.

## What changed
- Added `scripts/generate_coverage_report.mjs`
- Added npm commands:
  - `npm run coverage:extensions`
  - `npm run coverage:extensions:json`
- Updated README with usage and mapping rule

## How to run
```bash
npm run coverage:extensions
npm run coverage:extensions:json
```

Fixes #10

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,26 @@ Open `http://localhost:8080`.
   - `src/risc_v_visualizer.jsx` (`extensionInstructions`)
 - Sync script (merges `instr_dict.json` into `riscv_extensions.json`):
   - `scripts/sync_instructions.mjs`
+- Coverage report script (mapped/unmapped + group-wise summary):
+  - `scripts/generate_coverage_report.mjs`
+
+## Coverage report
+
+Run the report:
+
+```bash
+npm run coverage:extensions
+```
+
+Generate machine-readable JSON:
+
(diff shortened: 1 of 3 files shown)


amkr6207:
Hi @rpsene,

could you please review this PR? Thanks!

rpsene:

PR #24, opened 2026-04-28
Fix: Remove duplicate Zclsd extension and add data catalog validation

### LINKED ISSUE #22: Bug: Duplicate Extension ID "Zclsd" Causes UI and Data Inconsistencies
### Description
The extension `Zclsd` (Compressed LS-Pair) currently appears twice in the extension catalog data. It exists in both the `z_compress` and `z_system` groups:
- `z_compress` group (line 634): `{ id: 'Zclsd', ... desc: 'Compressed LS-Pair' }`
- `z_system` group (line 776): `{ id: 'Zclsd', ... desc: 'Compressed LS Pair' }`
This duplication is present in both `src/riscv_extensions.json` and the hardcoded fallback data in `src/risc_v_visualizer.jsx`.
### Impact
**MEDIUM** — This duplication causes several downstream issues:
1. The `Zclsd` extension tile renders twice in the UI.
2. The sync script writes instruction data to both entries.
3. Profile and Volume membership checks behave unpredictably (they often rely on first-match).
4. React `key` warnings appear in the console when rendered in the same list.
### Proposed Fix
We need to remove the duplication and enforce uniqueness:
- Remove the duplicate entry from the `z_system` group in `src/riscv_extensions.json` (its canonical group should be `z_compress`).
- Remove the corresponding duplicate from the fallback data in `src/risc_v_visualizer.jsx`.
- Update `scripts/sync_instructions.mjs` to include a validation step that detects duplicate IDs across groups and throws a warning or error.
- Re-run the sync script to correctly update the catalog data.

ARYANPATEL-BIT:
### Description
This PR resolves Issue #22 by removing the duplicate `Zclsd` extension from the application's catalog and adding validation to prevent similar issues from being introduced in the future.
Having duplicate IDs caused several problems:
1. React `key` warnings due to duplicate list items.
2. The `Zclsd` extension tile rendering twice in the UI.
3. Unpredictable first-match behavior when checking Profile/Volume memberships.
### Changes Made
- **Removed Duplicate Data**: Cleaned up `src/risc_v_visualizer.jsx` by removing the incorrect `Zclsd` entry from the `z_system` fallback data (its canonical location is `z_compress`). Verified that `src/riscv_extensions.json` is already correct.
- **Added Sync Validation**: Updated `scripts/sync_instructions.mjs` to validate the `extensionsCatalog` when building the index. The script now keeps track of extension IDs and throws an explicit `ERROR` if any duplicate ID is found across groups, failing the build process gracefully.
### Closes
Closes #22
### Verification
- [x] Verified `Zclsd` no longer appears twice in the UI.
- [x] Ran `node scripts/sync_instructions.mjs` successfully and it no longer threw any errors.
- [x] Verified successful production build via `npm run build`.

### DIFF
--- a/scripts/sync_instructions.mjs
+++ b/scripts/sync_instructions.mjs
@@ -82,6 +82,7 @@ function extractExtensionInstructions(jsxText) {
 
 function buildExtensionIndex(extensionsCatalog) {
   const index = new Map();
+  const duplicates = new Set();
 
   for (const [category, entries] of Object.entries(extensionsCatalog)) {
     if (!Array.isArray(entries)) continue;
@@ -92,7 +93,15 @@ function buildExtensionIndex(extensionsCatalog) {
       const list = index.get(id) ?? [];
       list.push({ category, entry });
       index.set(id, list);
+      
+      if (list.length > 1) {
+        duplicates.add(id);
+      }
     }
+  }
+
+  if (duplicates.size > 0) {
+    die(`ERROR: Duplicate extension IDs found across groups: ${Array.from(duplicates).join(', ')}`);
   }
 
   return index;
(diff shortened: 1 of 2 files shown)


rpsene:

PR #26, opened 2026-04-28
fix: Clean up event listeners and timeouts on unmount

### LINKED ISSUE #25: Memory Leak — Event Listeners and Timeouts Not Cleaned Up on Unmount
## Description

The `EncodingDiagram` and `RISCVExplorer` components in `src/risc_v_visualizer.jsx` have two cleanup issues that cause **"setState on unmounted component"** warnings and minor memory leaks:

### 1. RAF Race Condition (`EncodingDiagram`, lines 403–426)
The `requestAnimationFrame` cleanup nullifies `rafRef.current` **before** calling `cancelAnimationFrame`. If the cleanup runs between the RAF scheduling and its callback firing, the callback invokes `updateScrollState()` → `setScrollState()` on an unmounted component.

### 2. Uncleared `setTimeout` Calls (`RISCVExplorer`, lines ~3118, ~3487)
The copy-to-clipboard handlers use:
```js
window.setTimeout(() => setCopyStatus(null), 1500);
window.setTimeout(() => setEncoderValidatorCopyStatus(null), 1500);
```
without storing or clearing the timeout IDs. If the component unmounts within 1.5s of a copy action, the callbacks fire against stale state.

## Steps to Reproduce

1. Open the RISC-V Extensions Landscape app
2. Select any instruction → click **Copy** to trigger the copy-status timeout
3. Navigate away or unmount the component within 1.5 seconds (easily triggered in React Strict Mode)
4. Observe console warning: _"Can't perform a React state update on an unmounted component"_

## Impact

**MEDIUM** — Potential "setState on unmounted component" warnings and minor memory leaks in React Strict Mode or fast-navigation scenarios.

## Proposed Fix

**Branch:** `fix/cleanup-event-listeners-on-unmount`  
**Estimated change:** ~35–40 lines

1. **Fix RAF cleanup order** — snapshot the RAF ID before nullifying the ref so `cancelAnimationFrame` always receives the correct handle
2. **Add a `mountedRef`** — `useRef(true)` flag set to `false` on unmount
3. **Store timeout IDs in refs** (`copyTimeoutRef`, `encoderCopyTimeoutRef`) and clear them in the cleanup function
4. **Guard all `setTimeout` callbacks** — only call `setState` if `mountedRef.current` is `true`

ARYANPATEL-BIT:
…on unmount## Summary

Fixes #25 — Resolves memory leaks and "setState on unmounted component" warnings caused by uncleared event listeners and timeouts in `EncodingDiagram` and `RISCVExplorer`.

#### 1. EncodingDiagram — RAF race condition fix
The cleanup was nullifying `rafRef.current` before calling `cancelAnimationFrame`, creating a window where the RAF callback could fire on an unmounted component. Fixed by snapshotting the ID first:
```diff
-      if (rafRef.current) cancelAnimationFrame(rafRef.current);
-      rafRef.current = null;
+      const pendingRaf = rafRef.current;
+      rafRef.current = null;
+      if (pendingRaf) cancelAnimationFrame(pendingRaf);
```

#### 2. RISCVExplorer — Mounted guard + timeout cleanup
- Added `mountedRef` to track component lifecycle
- Added `copyTimeoutRef` and `encoderCopyTimeoutRef` to store timeout IDs
- Cleanup effect clears all pending timeouts on unmount

#### 3. Copy-status setTimeout calls (×2)
Both `setCopyStatus` and `setEncoderValidatorCopyStatus` timeouts now:
- Store the timeout ID in a ref
- Clear any previous timeout before setting a new one
- Guard the `setState` call with `if (mountedRef.current)`

## Testing

- ✅ Production build passes (zero errors)
- ✅ Copy button still works correctly with 1.5s status reset
- ✅ Fast unmount no longer triggers React warnings

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -420,8 +420,12 @@ const EncodingDiagram = ({ encoding }) => {
     return () => {
       el.removeEventListener('scroll', onScroll);
       window.removeEventListener('resize', onResize);
-      if (rafRef.current) cancelAnimationFrame(rafRef.current);
+      // Cancel any pending RAF *before* nullifying the ref to avoid
+      // a race where the callback fires between null-assignment and
+      // cancelAnimationFrame, calling setState on an unmounted component.
+      const pendingRaf = rafRef.current;
       rafRef.current = null;
+      if (pendingRaf) cancelAnimationFrame(pendingRaf);
     };
   }, [updateScrollState, normalized]);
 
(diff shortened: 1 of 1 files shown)


rpsene:

PR #32, opened 2026-05-01
fix: add missing PAUSE instruction for Zihintpause extension

RiteshBiradar:
### Problem
Zihintpause extension exists in the catalog but has no instruction mapping.
As a result, no instruction data is shown in the UI.

### Fix
Added missing mapping in `extensionInstructions`:
Zihintpause → ['PAUSE']

### Result
Instruction now appears correctly when selecting Zihintpause in the UI.

This is one of several extensions missing mappings. Similar fixes can be applied for other extensions.

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -1247,6 +1247,9 @@ const RISCVExplorer = () => {
       // Instruction-fetch fence
       'FENCE.I',
     ],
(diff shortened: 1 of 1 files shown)


rpsene:
Closing — the commits here are contained in #33, so merging that lands this work too. Verified with `git merge-base --is-ancestor`; nothing is lost and your commit authorship stays intact. Context: #138

PR #33, opened 2026-05-01
Fix/show missing instruction fallback

RiteshBiradar:
### Problem
Some extensions exist in the catalog but do not have instruction mappings.
When such an extension is selected, the UI shows an empty section, which looks broken and provides no feedback.

### Example
Clicking on extensions like `Zinctr`, `Zicntrpmf`, `Zibi` results in no instruction data being displayed.

### Root Cause
The UI renders instructions only when `extensionInstructions[selectedExt.id]` exists.
If it is undefined, nothing is rendered.

### Fix
Added a fallback message when instruction data is missing:
> "No instruction data available for this extension"

### Result
- Prevents empty UI state
- Provides clear feedback to users
- Makes incomplete mappings visible

### Notes
Previously fixed a related issue where `Zihintpause` was missing its instruction mapping (`PAUSE`).
This PR improves the experience for other unmapped extensions.

Future improvement: visually highlight incomplete extensions in the UI.

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -3021,7 +3021,7 @@ const RISCVExplorer = () => {
 	                      </div>
 	                    )}
 
-	                  {extensionInstructions[selectedExt.id] && (
+	                  {extensionInstructions[selectedExt.id] ? (
 	                    <div className="bg-slate-900 p-3 rounded border border-slate-700">
 	                      <h4 className="text-[10px] uppercase tracking-wider text-emerald-400 font-bold mb-2">
 	                        Instruction Set Snapshot ({extensionInstructions[selectedExt.id].length})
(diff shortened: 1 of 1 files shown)


rpsene:
Thanks, and apologies for the delay.

Closing because this cannot be rebased. It edits `extensionInstructions`, a hand-maintained array in the JSX, and that structure no longer exists: instruction data now lives in src/instr_dict.json and src/riscv_extensions.json, populated by scripts/sync_instructions.mjs. There are zero references to `extensionInstructions` left in src/. That change also closed #34, which was about the vm.runInNewContext extraction it required.

On the substance, the fallback message for an extension with no instructions needs care rather than a plain "none found". Around 125 of 227 entries correctly have no instructions: umbrellas and aliases such as K, N and P, VLEN parameter entries from Zvl32b to Zvl1024b, and behavioural guarantees like Zkt and Ztso which define no encodings at all. A message implying data is missing would be wrong for most of them.

Framed as "this extension defines no instructions" with the reason, it would be a good addition. Worth a fresh PR against the current structure if you are interested.

PR #36, opened 2026-05-01
perf: implement lazy loading for riscv_extensions.json

### LINKED ISSUE #35: Performance: Implement Lazy Loading for Large JSON Data
### Description
Currently, the `riscv_extensions.json` file (which is ~528KB) and other associated JSON data are statically imported and bundled into the main JavaScript bundle. They are parsed synchronously during initial module evaluation. 

This happens even if the user only views a few extensions, meaning they have to download and parse the entire catalog of 222 extensions up-front before the app becomes interactive.

### Proposed Fix
1. Ensure `instr_dict.json` is not imported into the visualizer (it should only be needed by the sync script via `fs`).
2. Use dynamic `import()` for `riscv_extensions.json` to load the data asynchronously after the initial component mounts.
3. Add a loading skeleton or spinner to the UI while the data fetches.
4. Ensure Webpack is configured to emit the JSON as a separate chunk.

ARYANPATEL-BIT:
This PR addresses the performance bottleneck caused by synchronously parsing the large `riscv_extensions.json` file (~528KB) on initial load. 

Previously, the massive extension catalog was statically imported and bundled directly into the main JavaScript chunk, blocking the initial render even if the user only needed to view a few extensions. This PR implements dynamic lazy loading to resolve the issue.

### Changes Made
- **Removed Static Import**: Removed the static `import extensions from './riscv_extensions.json'` inside `src/risc_v_visualizer.jsx`.
- **Dynamic Import with `useEffect`**: Added React state and a `useEffect` hook to dynamically fetch the JSON file (`import()`) only after the application mounts.
- **Loading State UI**: Added a sleek loading spinner and fallback text that is displayed to the user while the JSON payload is being fetched over the network.
- **Webpack Chunking**: Added `chunkFilename: '[name].[contenthash].js'` to `webpack.config.js` to explicitly separate dynamic imports into clear, cacheable chunks.

### Impact
- **Significant Bundle Size Reduction:** The 500KB+ JSON file is no longer baked into `bundle.js`, drastically reducing the initial JavaScript payload size.
- **Improved Time-To-Interactive (TTI):** Faster parsing and execution of the core app bundle on slower devices and low-bandwidth connections.

### Verification
- Tested locally by building the app with Webpack and verifying that the `riscv_extensions.json` data is correctly emitted as a separate chunk.
- Verified that the fallback loading spinner renders properly and seamlessly transitions once the data is resolved.

Fixes: #35

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -10,7 +10,6 @@ import {
   ChevronLeft,
   ChevronRight,
 } from 'lucide-react';
-import extensions from './riscv_extensions.json';
 
 const BIT_WIDTH = 32n;
 const BIT_MASK_32 = (1n << BIT_WIDTH) - 1n;
@@ -564,6 +563,7 @@ const EncodingDiagram = ({ encoding }) => {
 };
 
 const RISCVExplorer = () => {
+  const [extensions, setExtensions] = useState(null);
   const [activeProfile, setActiveProfile] = useState(null);
   const [activeVolume, setActiveVolume] = useState(null);
   const [selectedExt, setSelectedExt] = useState(null);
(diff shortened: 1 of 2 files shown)


rpsene:

PR #38, opened 2026-05-01
fix: reject Match/Mask hex inputs wider than 32 bits in Encoder Validator

### LINKED ISSUE #37: Encoder Validator silently truncates Match/Mask hex inputs wider than 32 bits
### Description

The Encoder Validator accepts hex values wider than 32 bits and AND-masks them with `BIT_MASK_32` before running its checks. The user sees a conflict report based on the truncated value, with no indication the high bits were dropped. `parseHexToBigInt` in `src/risc_v_visualizer.jsx` does not bound-check the parsed value, and the "Match contains bits outside Mask" guard at line 2240 runs on the already-truncated values, so it cannot catch oversized input.

### Steps to reproduce

1. Open the Encoder Validator.
2. Enter `Match = 0x11800202f` and `Mask = 0xf800707f` (one extra hex digit on Match).
3. Click Validate.

The result lists a conflict against `SC.W` (whose real match is `0x1800202f`), even though that is not what was typed.

### Expected behavior

Reject inputs wider than 32 bits with a clear error like `Match exceeds 32 bits.`, and do not run conflict detection against a truncated value.

### Actual behavior

Oversized input is accepted, AND-masked to 32 bits, and run through conflict detection. The user sees results that do not correspond to what they typed.

### Code reference

- `src/risc_v_visualizer.jsx:282-291` — `parseHexToBigInt` does not check `value > BIT_MASK_32`.
- `src/risc_v_visualizer.jsx:2238-2242` — the "Match contains bits outside Mask" check runs on `matchNorm`/`maskNorm`, which are already truncated, so it cannot detect oversized input.

deepak0x:
## Proposed changes

The Encoder Validator was accepting hex Match/Mask values wider than 32 bits, AND-masking them with `BIT_MASK_32`, and reporting conflicts against the truncated value with no warning. This change adds two explicit width checks in the validator block so oversized input is rejected with `Match exceeds 32 bits.` / `Mask exceeds 32 bits.` instead of being silently coerced.

## Issue(s)

Closes #37

## How to test or reproduce

1. Open the Encoder Validator.
2. Enter `Match = 0x11800202f` and `Mask = 0xf800707f` (one extra hex digit on Match).
3. Click Validate.

Before this change: the validator returns a conflict against `SC.W` (whose real match is `0x1800202f`).
After this change: the validator returns the error `Match exceeds 32 bits.` and runs no conflict detection.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [x] Lint and unit tests pass locally with my changes (no eslint config / no test suite in repo; `npm run build` succeeds with only pre-existing bundle-size warnings)

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -2233,6 +2233,8 @@ const RISCVExplorer = () => {
       const maskParsed = parseHexToBigInt(proposedMaskInput);
       if (matchParsed == null) errors.push('Match must be a hex value like 0x1234.');
       if (maskParsed == null) errors.push('Mask must be a hex value like 0x707f.');
+      if (matchParsed != null && matchParsed > BIT_MASK_32) errors.push('Match exceeds 32 bits.');
+      if (maskParsed != null && maskParsed > BIT_MASK_32) errors.push('Mask exceeds 32 bits.');
 
       if (matchParsed != null && maskParsed != null) {
         const matchNorm = matchParsed & BIT_MASK_32;


deepak0x:
@rpsene could you take a look when you get a chance? Thanks!

rpsene:
Thanks, and sorry this sat so long. The same fix landed in #169, which closed #37.

The guard ended up at the validation site rather than inside parseHexToBigInt, because that helper also parses catalogue match/mask values, which are trusted and already 32-bit, so bounding it globally would have changed data parsing too. `0x11800202f` is now rejected with "Match exceeds 32 bits." instead of being truncated to `0x1800202f` and reported as a conflict against SC.W.

Closing as superseded. Your diagnosis was right.

PR #41, opened 2026-05-01
fix: clear stale search-driven selection when query has no match

### LINKED ISSUE #39: Selected Details panel stays open with stale extension after search query becomes non-matching
### Description

The search `useEffect` in `src/risc_v_visualizer.jsx` only clears state when the query is empty. When the query is non-empty but matches nothing, the previously search-opened extension and instruction panel remain visible, making it look like the new query matched.

### Steps to reproduce

1. Load the app.
2. Type `ADD` in the search box. The page scrolls to the RV32I tile and the Selected Details panel opens with the `ADD` instruction.
3. Extend the query to `ADDqqq` (or any string with no match).

Observed: the RV32I tile is still expanded as the selected extension and `ADD` is still shown as the selected instruction, even though the current query does not match RV32I, ADD, or any other extension.

Expected: when the query no longer matches anything, the search-driven selection clears (or at least the visible selection no longer points at something the query does not match).

### Code reference

`src/risc_v_visualizer.jsx:2452-2520` — the effect updates state inside `if (targetExt) { ... }` but has no `else` branch to clear state when the query is non-empty and `targetExt` ends up null after all three lookups (id, mnemonic, deep search index).

deepak0x:
## Proposed changes

The search `useEffect` in `src/risc_v_visualizer.jsx` previously cleared the search-driven selection only when the query was empty. If the query was non-empty but matched nothing (e.g. typing past a match like `ADD` → `ADDqqq`), the prior `selectedExt`, `selectedInstruction`, and `searchMatches` stayed in place, making the panel look like the new query had matched. This PR adds the missing else branch so a non-matching, non-empty query clears the same state and resets the scroll-tracking ref.

## Issue(s)

Closes #39

## How to test or reproduce

1. `npm install && npm run build && python3 -m http.server 8080 -d dist`
2. Open `http://localhost:8080`.
3. Type `ADD` in the search box — the page scrolls to the RV32I tile and the Selected Details panel opens with `ADD`.
4. Extend the query to `ADDqqq`.
5. Before this PR: the RV32I tile remains selected and `ADD` still shows in the Selected Details panel.
6. After this PR: the selection clears as soon as the query no longer matches anything.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read the CONTRIBUTING doc
- [x] Lint and unit tests pass locally with my changes

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -2514,6 +2514,11 @@ const RISCVExplorer = () => {
         }
 	        lastScrolledKeyRef.current = key;
 	      }
+	    } else {
+	      setSelectedExt(null);
+	      setSelectedInstruction(null);
+	      setSearchMatches(null);
+	      lastScrolledKeyRef.current = null;
 	    }
 	  }, [searchQuery, extensionSearchIndexById]);
 


deepak0x:
@rpsene whenever you have a moment, would appreciate a look at this — small, contained fix for the search-selection stale state. Happy to address any feedback. Thanks!

rpsene:

PR #45, opened 2026-05-02
fix: resolve conflicting profile and volume filter UI state

### LINKED ISSUE #44: Conflicting UI State: Profile and Volume Filters Can Be Active Simultaneously
### Description
Currently, the Profile filter (RVA20, RVA22, etc.) and the Volume filter (Vol I, Vol II) can be activated at the same time, leading to a conflicting UI state. 

When both are active:
- The `isHighlighted` function returns `true` if **EITHER** the profile **OR** volume matches.
- However, the `isDimmed` function only checks the profile, returning `false` whenever a volume is active.
This creates a very confusing interaction:
1. Selecting the RVA20 profile correctly dims non-RVA20 extensions.
2. Then, selecting Vol II un-dims everything (because `isDimmed` returns `false` when `activeVolume` is set).
3. The yellow highlight ring appears on extensions matching EITHER filter, with no visual indication of which filter is causing the highlight.

### Proposed Solution
We need to handle the state and visual representation more cleanly. 
- [ ] Make profile and volume filters mutually exclusive:
  - When a profile is selected, clear `activeVolume`.
  - When a volume is selected, clear `activeProfile`.
- [ ] Update the `isDimmed` logic to properly handle the new state.
- [ ] Add a "Clear filters" button that resets both active states.

ARYANPATEL-BIT:
### Description
This PR addresses the conflicting UI state that occurs when the Profile and Volume filters are active simultaneously. Previously, having both active caused overlapping highlights without clear indications of which filter caused them, and the dimming logic was broken.
### Changes Made
- **Mutually Exclusive State:** Updated the `onClick` handlers. Selecting a Profile now explicitly clears any active Volume (and vice versa).
- **Corrected Dimming Logic:** Updated `isDimmed` to properly handle either filter state instead of prematurely returning `false` when a Volume was active.
- **Distinct Highlight Colors:** To improve the visual experience, `ExtensionBlock` now uses distinct highlight rings:
  - **Blue** for Profile matches
  - **Green** for Volume matches
  - **Yellow** remains the default for search matches and selections
- **Clear Filters Button:** Added a new "Clear Filters" button next to the Volume buttons. This button dynamically appears when any filter is active and provides a quick way to reset the UI.
### How to Test
1. Click on a Profile (e.g., RVA20) — verify the ring turns blue and non-matching extensions dim.
2. Click on a Volume (e.g., Vol I) — verify the Profile is cleared, the Volume ring turns green, and dimming updates correctly.
3. Verify the "Clear Filters" button appears when a filter is active.
4. Click "Clear Filters" and verify all highlights/dims are removed.

Fixes: #44

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -2385,9 +2385,9 @@ const RISCVExplorer = () => {
   };
 
   const isDimmed = (id) => {
-    if (activeVolume) return false;
-    if (!activeProfile) return false;
-    return !profiles[activeProfile].includes(id);
+    if (activeVolume) return !volumeMembership[activeVolume]?.has(id);
+    if (activeProfile) return !profiles[activeProfile].includes(id);
+    return false;
   };
 
   const ExtensionBlock = ({ data, colorClass, searchQuery }) => {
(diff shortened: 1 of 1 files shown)


rpsene:

PR #48, opened 2026-05-03
fix: debounce search input to prevent excessive re-renders

### LINKED ISSUE #47: Performance: Debounce search input to prevent excessive re-renders and lag
**Description:**

### The Problem
Currently, the search `useEffect` runs on every single keystroke (firing whenever `searchQuery` changes). 

For each keystroke, the application:
1. Iterates through all ~222 extensions to find matches.
2. Iterates through potentially thousands of instructions within the matched extensions.
3. Updates `selectedExt`, `searchMatches`, and `selectedInstruction` states.
4. Triggers a DOM `scrollIntoView`.

Because this runs $O(n \times m)$ matching on every keystroke with no debouncing, the state updates cause the entire component tree to re-render constantly. 

### Impact
Users experience perceived input lag on slower devices, excessive unnecessary re-renders, and janky scroll behavior during fast typing.

### Proposed Solution
1. **Add a debounce mechanism (e.g., 300ms)** before executing the search logic:
   ```javascript
   const [debouncedQuery, setDebouncedQuery] = useState('');
   
   useEffect(() => {
     const timer = setTimeout(() => setDebouncedQuery(searchQuery), 300);
     return () => clearTimeout(timer);
   }, [searchQuery]);
   ```
2. Move the heavy search `useEffect` to depend on `debouncedQuery` instead of the raw `searchQuery`.
3. *(Optional but recommended)* Implement a "Searching..." visual indicator during the debounce period.
4. *(Optional)* Consider using `useMemo` or a precomputed index to speed up lookups.

### Location
`src/risc_v_visualizer.jsx` (around lines 2450-2518)

ARYANPATEL-BIT:
**Description:**

Fixes a performance issue where the search `useEffect` was running on every single keystroke. 

Previously, the app would iterate through ~222 extensions and potentially thousands of instructions to update the state and trigger DOM scrolling on every keystroke, leading to significant perceived lag on slower devices.

### Changes made
- Added a `debouncedQuery` state to delay search execution.
- Implemented a `setTimeout` of 300ms to update the debounced state only after the user stops typing.
- Updated the primary search `useEffect` dependency to run off `debouncedQuery` instead of the raw `searchQuery`.
- Added a visual `"Searching..."` indicator below the search bar that displays during the 300ms debounce period to improve UX.

### Related Issues
Resolves #47 

### Testing
- [x] Verified that typing rapidly does not cause the app to stutter.
- [x] Verified that the `"Searching..."` text appears while typing and disappears when search executes.
- [x] Verified that the correct extension is highlighted and scrolled into view once the debounce timer finishes.

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -570,6 +570,13 @@ const RISCVExplorer = () => {
   const [selectedInstruction, setSelectedInstruction] = useState(null);
   const [copyStatus, setCopyStatus] = useState(null);
   const [searchQuery, setSearchQuery] = useState('');
+  const [debouncedQuery, setDebouncedQuery] = useState('');
+
+  React.useEffect(() => {
+    const timer = setTimeout(() => setDebouncedQuery(searchQuery), 300);
+    return () => clearTimeout(timer);
+  }, [searchQuery]);
+
   const [searchMatches, setSearchMatches] = useState(null);
   const [encoderValidatorOpen, setEncoderValidatorOpen] = useState(false);
   const [encoderValidatorInput, setEncoderValidatorInput] = useState({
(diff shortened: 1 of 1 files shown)


rpsene:

Issue #56, opened 2026-05-04
Encoder Validator Modal Has No Escape Key or Focus Management


ARYANPATEL-BIT:
### Description
The encoder validator modal currently suffers from several accessibility and usability issues related to keyboard navigation:
- It cannot be closed using the `Escape` key (users must click the X button or the backdrop).
- It does not trap focus within the modal (`Tab` key can reach elements behind the backdrop).
- It does not return focus to the trigger button when closed.
- It is missing essential accessibility attributes (`aria-modal="true"` and `role="dialog"`).
- The backdrop click handler closes the modal, but there's no keyboard equivalent for non-mouse users.

### Proposed Fix
- [ ] Add an `onKeyDown` handler to the modal container to support the `Escape` key.
- [ ] Implement focus trapping:
  - On open, move focus to the first input in the modal.
  - On `Tab` at the last focusable element, wrap to the first.
  - On `Shift+Tab` at the first focusable element, wrap to the last.
  - On close, return focus to the "Encoder Validator" button.
- [ ] Add `role="dialog"` and `aria-modal="true"` to the modal container.
- [ ] Add `aria-labelledby` pointing to the modal title.
- [ ] Add `aria-describedby` pointing to the modal description.

**Location:** `src/risc_v_visualizer.jsx`

PR #61, opened 2026-05-05
fix: add a11y focus management to encoder modal

ARYANPATEL-BIT:
### Description
This PR resolves an accessibility and usability issue with the Encoder Validator modal where users relying on keyboard navigation were unable to interact with or exit the modal efficiently. 

**Resolves Issue:** #56 

### Changes Made
- **Escape Key Support:** Added an `onKeyDown` listener to allow closing the modal using the `Escape` key.
- **Focus Trapping:** Intercepted `Tab` and `Shift+Tab` keys to keep keyboard focus strictly within the dialog while it is open. 
- **Auto-Focus:** Focus is now programmatically shifted to the first input field as soon as the modal is opened.
- **Focus Restoration:** The modal now remembers the trigger element that opened it and restores focus to that element when closed.
- **ARIA Attributes:** Added `role="dialog"`, `aria-modal="true"`, `aria-labelledby`, and `aria-describedby` to the modal container for proper screen reader support.

### How to Test
1. Open the application and navigate to the Encoder Validator button.
2. Open the modal using the `Enter` or `Space` key.
3. Verify that focus jumps immediately to the first input field.
4. Press `Tab` repeatedly and verify that focus loops back to the top when reaching the end of the modal (focus does not escape to the background).
5. Press `Escape` and verify the modal closes.
6. After closing, verify that focus is automatically returned to the Encoder Validator trigger button.

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -582,6 +582,70 @@ const RISCVExplorer = () => {
   const [encoderValidatorCopyStatus, setEncoderValidatorCopyStatus] = useState(null);
   const lastScrolledKeyRef = React.useRef(null);
 
+  const modalRef = React.useRef(null);
+  const previousFocusRef = React.useRef(null);
+
+  React.useEffect(() => {
+    if (encoderValidatorOpen) {
+      previousFocusRef.current = document.activeElement;
+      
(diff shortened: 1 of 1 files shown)


rpsene:

PR #62, opened 2026-05-05
feat(data): map instruction sets for vector crypto bundle extensions

Divyateja2709:
Description

This PR adds instruction mappings for vector crypto bundle extensions in the z_vector_crypto category.

These extensions are not standalone instruction sets; they are defined in the RISC-V Vector Crypto specification as compositions of existing sub-extensions. Their instruction sets are therefore implemented as the union of their respective components.

The following bundle extensions are added:

Zvkb = Zvbb + Zvbc
Zvkng = Zvkn + Zvkg
Zvknc = Zvkn + Zvbc
Zvksg = Zvks + Zvkg
Zvksc = Zvks + Zvbc

Each extension has been mapped by aggregating instructions from its already-supported base extensions, ensuring consistency with the spec and avoiding duplication of logic.

Changes Made
Updated src/risc_v_visualizer.jsx to include:
Zvkb
Zvkng
Zvknc
Zvksg
Zvksc
Populated instruction sets as unions of corresponding sub-extensions
Ran node scripts/sync_instructions.mjs to sync generated data

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,7 +73,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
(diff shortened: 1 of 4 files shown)


rpsene:
Closing — the commits here are contained in #64, so merging that lands this work too. Verified with `git merge-base --is-ancestor`; nothing is lost and your commit authorship stays intact. Context: #138

PR #63, opened 2026-05-05
feat(data): improve metadata for Zvl and Zve extensions

Divyateja2709:
Description

This PR enhances the metadata for vector length (Zvl*) and vector subset (Zve*) extensions in riscv_extensions.json.

Unlike standard extensions, these do not define instruction sets. Instead, they act as capability markers that describe implementation constraints such as minimum vector length (VLEN) and supported subsets of vector functionality.

This update clarifies their role to avoid confusion during instruction mapping and visualization.

Changes Made
Updated desc fields for:
Zvl32b → Zvl1024b (vector length constraints)
Zve32x → Zve64d (vector subset profiles)
Updated use fields to clearly explain:
These extensions do not introduce opcodes
They define hardware capabilities and constraints
Their relationship with the base vector extension (V)
Added consistent explanatory language across all Zvl* and Zve* entries


rpsene:
Closing — the commits here are contained in #64, so merging that lands this work too. Verified with `git merge-base --is-ancestor`; nothing is lost and your commit authorship stays intact. Context: #138

PR #64, opened 2026-05-05
feat(data): map hint and ordering extension instructions (Zihintpause, Zihintntl, Ztso)

Divyateja2709:
Description

This PR adds instruction mappings for the hint and memory ordering extensions:

Zihintpause
Zihintntl
Ztso

These extensions are unique in that they do not introduce new opcodes. Instead, they define semantic hints or ordering guarantees using existing instruction encodings.

This update ensures they are properly represented in the visualizer while clearly documenting their special nature.

Changes Made
Updated src/risc_v_visualizer.jsx to include:
Zihintpause
Added:
PAUSE
Documented that it is encoded via FENCE with specific operands
Zihintntl
Added:
NTL.P1, NTL.PALL, NTL.S1, NTL.ALL
Documented that these reuse existing ADDI/ADD encodings
Ztso
Added explanatory handling (if applicable in your codebase)
Clarified that it enforces Total Store Ordering (TSO) semantics rather than defining instructions
Added inline comments explaining:
These extensions reuse existing encodings
They function as hints or memory model constraints, not new instruction sets


rpsene:
Thanks for this, and the analysis was right.

Closing the PR because it edits `extensionInstructions`, which no longer exists: instruction data moved to src/instr_dict.json and src/riscv_extensions.json, and there are zero references left in src/. So it cannot be rebased, only rewritten.

The finding itself is real and I did not want it to disappear with the branch, so I have opened #178 to carry it. Your table of Zihintpause, Zihintntl and Ztso is quoted there, along with the reason it was not done alongside the bitmanip work in #170: pseudo-instructions share an encoding with the instruction they alias, and rv32_zclsd in particular redefines c.ld and c.sd onto the C.FLW and C.FSW encodings, which collides with the existing RV64 mnemonics. That needs a decision about representing aliases before the data goes in.

You are credited in the issue. If you want to take it once the alias question is settled, it is yours.

PR #71, opened 2026-05-06
feat: add Zvkn and Zvks vector crypto instruction mappings

caffeine-rohit:
## What
Adds instruction mappings for two vector cryptography extensions 
that were missing from `extensionInstructions` in the visualizer:

- **Zvkn** (Vector NIST Suite) — 23 instructions: AES encrypt/decrypt, 
  SHA-2, and bitmanip (VANDN, VBREV8, VROL, VROR)
- **Zvks** (Vector ShangMi Suite) — 14 instructions: SM4, SM3, 
  and bitmanip

## Why
Both extensions existed in the catalog (`riscv_extensions.json`) 
and had full encoding data in `instr_dict.json`, but were not 
registered in `extensionInstructions`, so the sync script never 
populated them. They showed as empty in the UI.

## How to verify
```bash
node scripts/sync_instructions.mjs
```
Sync completes with zero warnings. Both extensions now show 
full instruction data in the landscape UI.

## Related
Part of the effort to map the remaining extensions that 
currently show no instruction data.

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -1737,6 +1737,20 @@ const RISCVExplorer = () => {
       // Vector GCM/GMAC
       'VGHSH.VV', 'VGMUL.VV',
     ],
+    Zvkn: [
+      // Vector NIST suite (AES + SHA + bitmanip)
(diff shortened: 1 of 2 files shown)


rpsene:

PR #79, opened 2026-05-08
use debouncedQuery in ExtensionBlock renders to eliminate per-keystroke grid re-renders

### LINKED ISSUE #51: Bug: Raw searchQuery forces all 222 ExtensionBlocks to re-render on every keystroke despite debounce
### Description
The debounce introduced in PR #48 delays the expensive search computation, but it does not prevent unnecessary DOM re-renders.

The raw `searchQuery` is still passed to all `<ExtensionBlock>` components. This causes the entire grid (222 items) to re-render instantly on every keystroke.

**Location:**  
`src/risc_v_visualizer.jsx`

```jsx
<ExtensionBlock
  key={item.id}
  data={item}
  searchQuery={searchQuery}
/>
```

---

###  Impact
- Full grid re-renders on each input change  
- Unnecessary reconciliation across all child components  
- Noticeable typing lag and UI slowdown  

---

### Expected Behavior
- Re-renders should be strictly driven by `debouncedQuery`, not raw `searchQuery`  
- `<ExtensionBlock>` should update only when relevant data changes  
- Typing should remain smooth without triggering full grid updates  

---

### Suggested Fix

```jsx
<ExtensionBlock
  key={item.id}
  data={item}
  searchQuery={debouncedQuery}
/>
```

**Optional optimization:**

```jsx
export default React.memo(ExtensionBlock);
```

Annu881:
## Description

Fixes a performance issue where all 222 `<ExtensionBlock>` components
re-rendered on every single keystroke during search input.

The raw `searchQuery` state was passed to every tile in the grid, forcing
React to reconcile the entire component tree on each input change — causing
noticeable typing lag and UI slowdown.

## Changes Made

- Replaced `searchQuery={searchQuery}` with `searchQuery={debouncedQuery}`
  across all `<ExtensionBlock>` usages in the extension grid
- Tile highlights now update only after the 300ms debounce settles,
  eliminating unnecessary per-keystroke re-renders

## Related Issues

Resolves #51

## Testing

- [x] Verified grid tiles no longer re-render on every keystroke
- [x] Verified tile highlights update correctly after debounce completes
- [x] Build compiles successfully with no errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,29 +4,31 @@
   "description": "",
   "main": "index.js",
   "homepage": "https://rpsene.github.io/riscv-extensions-landscape/",
-"scripts": {
-  "build": "webpack",
-  "predeploy": "npm run build",
-  "deploy": "gh-pages -d dist --branch gh-pages --dotfiles",
-  "test": "echo \"Error: no test specified\" && exit 1"
-},
+  "scripts": {
+    "dev": "webpack serve --mode development",
+    "build": "webpack",
+    "predeploy": "npm run build",
+    "deploy": "gh-pages -d dist --branch gh-pages --dotfiles",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
   "keywords": [],
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "gh-pages": "^6.0.0",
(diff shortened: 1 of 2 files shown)


rpsene:

PR #87, opened 2026-05-10
refactor: extract extensionInstructions to JSON, remove vm.runInNewContext

### LINKED ISSUE #34: Refactor `sync_instructions.mjs` to remove `vm.runInNewContext` security risk and extraction fragility
## Description
The current `scripts/sync_instructions.mjs` script extracts `extensionInstructions` from the main JSX file using a fragile string-matching and VM execution approach.

Currently, the script:
- Finds the string `const extensionInstructions =`
- Extracts the text between matching braces
- Runs the extracted text through `vm.runInNewContext()`

##  The Problem
This approach presents several maintainability and security concerns:

- **Security & Fragility:** It executes arbitrary JavaScript from the JSX file in a sandboxed VM context. While sandboxed, this is highly fragile and could execute unexpected code or break completely if the JSX format changes (e.g., if the object uses imports, function calls, or template literals).
- **Parsing Limitations:** It relies on a custom regex-like brace-matching algorithm that doesn't handle all JavaScript edge cases (such as regex literals containing braces, or multi-line template literals with nested expressions).
- **Escape Detection Bug:** The escape detection on line 25 uses `\\\\` which acts as a double backslash literal rather than a single backslash character check.

**Impact:** MEDIUM. The extraction process is fragile and will easily break as the data format evolves. It also poses a potential security concern if untrusted JSX is processed.

##  Proposed Solution
We should separate the data from the component logic to eliminate the need for AST parsing / VM execution.

1. Move the `extensionInstructions` data to a dedicated JSON or JS data file (e.g., `src/data/extension_instructions.json`).
2. Update `scripts/sync_instructions.mjs` to simply read and parse this file (`JSON.parse(fs.readFileSync(...))`) instead of evaluating it.
3. Update the visualizer component to import this new data file (`import extensionInstructions from './data/extension_instructions.json'`).
4. Remove the brittle brace-matching and `vm.runInNewContext` logic entirely.
5. Add proper error handling with descriptive console messages to the sync script.

praveen24sriv:
Closes #34

## What
- Moves `extensionInstructions` (~830 lines of inline data) out of `risc_v_visualizer.jsx` into `src/data/extension_instructions.json`
- Rewrites `sync_instructions.mjs` to read the JSON file directly via `JSON.parse` — eliminates `vm.runInNewContext` and the fragile `findMatchingBrace` parser entirely
- Updates `risc_v_visualizer.jsx` to import from the new data file

## Why
- `vm.runInNewContext` is not a true security sandbox and executes arbitrary JS extracted via string matching
- The hand-rolled brace parser fails on regex literals, nested template literals, and has an escape detection bug (line 25 uses `\\\\` instead of `\\`)
- Data and component logic should be separated

## Testing
- `node scripts/sync_instructions.mjs` → 1416 instruction entries, no errors
- `npm run build` → clean build, no new warnings


rpsene:
Thanks. This is resolved, though by a different route than proposed.

`extensionInstructions` no longer exists in the JSX at all. Instruction data lives in src/instr_dict.json and src/riscv_extensions.json, and sync_instructions.mjs reads those directly, so there is no vm.runInNewContext and no brace matching left to remove. I closed #34 on that basis.

Closing as superseded. The security concern was legitimate.

PR #92, opened 2026-05-14
feat: add canonical tag normalizer for automated sync

Rahulatram321:
Implements semantic tag normalization to unblock automated sync between riscv-opcodes and the landscape catalog.

- Adds scripts/lib/tag_normalizer.mjs with full JSDoc documentation

- Supports rv_/rv32_/rv64_ prefix stripping, lowercase normalization

- Handles G expansion (I+M+A+F+D+Zicsr+Zifencei)

- Includes comprehensive test suite

- Adds --use-normalizer flag to sync script (opt-in)

- Documents normalization rules in README

Based on analysis from PR #1.


rpsene:

PR #94, opened 2026-05-14
feat: add extension permalink sharing

Veekshitha11:
### What this adds
Two connected features:
1. **URL deep linking** — opening `?ext=Zba` pre-selects that extension on load
2. **Share button** — copies a permalink URL for the currently selected extension

### Why
There is currently no way to share a direct link to a specific extension.
This makes the tool significantly more useful in documentation and discussions.

### Changes in `src/risc_v_visualizer.jsx`
- Added `Share2` icon import from lucide-react
- Added module-level `allExtensions` array (replaces 3 duplicated inline flattening calls)
- Added `getExtensionById(id)` — case-insensitive lookup, returns null safely
- Added `getInitialSelectedExtension()` — reads `?ext=` on load via lazy useState init
- Added `buildExtensionPermalink(extId)` — builds a full shareable URL
- Added `shareStatus` state for button feedback (Share → Copied / Copy failed → Share)
- Added `copySelectedExtensionPermalink` useCallback
- Added Share button with `aria-label` in the extension detail panel header

### How to test
1. `npm run build && cd dist && python3 -m http.server 8080`
2. Open `http://localhost:8080/?ext=Zba` — Zba should be pre-selected
3. Click any extension → Share button appears → click it → paste clipboard → URL contains `?ext=<id>`
4. Open that URL — correct extension is pre-selected

### Note
Clicking extension tiles does not update the address bar live —
only the Share button generates the permalink. This is intentional scope.

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -7,6 +7,7 @@ import {
   ArrowRight,
   ArrowUpRight,
   Copy,
+  Share2,
   ChevronLeft,
   ChevronRight,
 } from 'lucide-react';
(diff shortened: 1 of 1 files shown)


rpsene:
Implemented and merged in #180, with the design credited to you.

Your approach carried over intact: the ?ext= query parameter, the case-insensitive lookup so ?ext=zba works as readily as ?ext=Zba, the lazy initial state so the deep link is honoured on first paint rather than selecting nothing and correcting itself, and the copy button.

I closed this rather than asking for a rebase because the branch conflicts with main and src/risc_v_visualizer.jsx has moved a long way since May, so rebasing would have been more work for you than the feature took to write.

Two things were added on top:

The address bar now tracks the selection rather than updating only when the button is pressed, so copying the URL straight from the browser works and a reload or bookmark returns to the same extension. It uses replaceState rather than pushState, since clicking through twenty tiles should not leave twenty history entries to walk back out of.

The parameter is canonicalised, so ?ext=zba resolves to ?ext=Zba and shared links look consistent however they were typed.

One note you may find interesting: your choice of a query parameter over a path turned out to matter beyond style. The site is static files on GitHub Pages with no router, so /extensions/Zba would 404 on a hard refresh without a server-side fallback. That is also why #120, which proposed nginx SPA routing, was closed as unnecessary.

Thank you for the contribution, and sorry it took this long to act on.

PR #95, opened 2026-05-14
docs: explain visualizer data flow helpers

Veekshitha11:
### What this adds
JSDoc comments above the five most important helper functions in
`risc_v_visualizer.jsx`, plus a small refactor that extracts one
inline async handler into a proper named callback.

### Why
The file is 3500+ lines. Key functions like the search matcher,
clipboard helper, and encoder validator had no documentation.
New contributors had no way to understand the data flow without
reading all the implementation code.

### Changes in `src/risc_v_visualizer.jsx`
- Added JSDoc above `instructionMatchesQuery` — explains 3-field search matching
- Added JSDoc above `copyTextToClipboard` — documents Clipboard API + textarea fallback
- Extracted inline `async onClick` on instruction Copy button into named `copySelectedInstructionDetails` useCallback with correct dependency array
- Added JSDoc above `copySelectedInstructionDetails`
- Added JSDoc above `runEncoderValidation`
- Replaced plain comment on search `useEffect` with JSDoc block

### Behavior change
None. The only code change is extracting an inline function into a
named callback — the logic is identical, the dependencies are explicit.

### How to verify
`npm run build` → zero errors (3 size warnings expected).
Click any extension → click any instruction → click Copy → paste → details present.

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -1985,6 +1985,10 @@ const RISCVExplorer = () => {
     };
   }, []);
 
+  /**
+   * Checks one instruction against the active search term, including the mnemonic,
+   * encoding metadata, and extension tags that are not visible in the tile label.
+   */
   const instructionMatchesQuery = (mnemonic, details, q) => {
     const needle = String(q || '').trim().toLowerCase();
     if (!needle) return false;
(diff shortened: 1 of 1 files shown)


rpsene:
Thanks, and sorry this waited.

Closing because src/risc_v_visualizer.jsx has been rewritten substantially since May, so the comments no longer line up with the code they describe. Among other things the file lost 328 lines of dead commented-out catalogue, the tile component moved to its own file, the search effect changed, and the CSR lookups moved to module scope.

The instinct is right and the file is better commented now as a result of that work. If specific helpers still read as unclear, point at them and I will document those directly.

PR #96, opened 2026-05-14
chore: add editorconfig

Veekshitha11:
### What this adds
A `.editorconfig` file at the repo root enforcing consistent
editor defaults for all contributors.

### Why
Without it, different editors produce different formatting:
tabs vs spaces, CRLF vs LF, trailing whitespace, missing final
newlines. This creates noisy diffs and inconsistent style over time,
especially as the contributor count grows.

### Settings applied
| Setting | Value | Reason |
|---|---|---|
| charset | utf-8 | Matches existing source files |
| end_of_line | lf | Matches existing source files |
| indent_style | space | Matches existing .jsx code style |
| indent_size | 2 | Matches existing .jsx code style |
| insert_final_newline | true | Standard Unix convention |
| trim_trailing_whitespace | true | Clean diffs |

### Files changed
`.editorconfig` — new file, 9 lines

### No behavior change
This file has zero effect on the built application.
It only affects how editors save files on disk.

### Follow-up suggestion
A `.gitattributes` file with `* text=auto eol=lf` would reinforce
these settings at the Git level and prevent CRLF commits from
Windows machines where `core.autocrlf=true` is the default.


rpsene:
Thanks for this, and sorry it sat so long. I have reviewed it and approved the workflow run, and the change itself is fine: 9 lines of standard editor settings, consistent with the repo (2-space, lf, utf-8, final newline, trim trailing). It complements Prettier rather than duplicating it, since .editorconfig governs the editor and Prettier governs formatting on demand.

CI is green on build-and-test. The one thing blocking the merge is the DCO check, which we added last week and which your commit predates. Everything now needs a Signed-off-by line, the same mechanism the kernel and Linux Foundation projects use. There is no CLA.

One command from your branch:

```
git commit --amend -s
git push --force-with-lease
```

The name and email in the sign-off need to match the commit author. Full details are in CONTRIBUTING.md and the DCO file at the repo root.

Ping me once that is pushed and I will merge it straight away.

PR #97, opened 2026-05-14
Improve sync script warning summaries and reporting

AmitShah13:
## Summary

Improved the logging and warning output of `sync_instructions.mjs` to make synchronization results easier to inspect and debug.

## Changes
- Added structured sync summary output
- Added counts for missing extensions and instructions
- Improved formatting and readability of warnings
- Display actual missing instruction names instead of only counts
- Added final sync completion message

## Why
The previous warning output was difficult to scan when debugging synchronization and extension mapping issues. These improvements make it easier to identify missing mappings and instruction gaps during sync operations.

### DIFF
--- a/scripts/sync_instructions.mjs
+++ b/scripts/sync_instructions.mjs
@@ -144,14 +144,46 @@ for (const [extId, mnemonics] of Object.entries(extensionInstructions)) {
 
 fs.writeFileSync(catalogPath, `${JSON.stringify(extensionsCatalog, null, 2)}\n`);
 
-console.log(`Updated ${path.relative(workspaceRoot, catalogPath)} with ${addedCount} instruction entries.`);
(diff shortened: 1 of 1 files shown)


rpsene:
Thanks. Closing as largely covered.

The sync already prints a per-run summary: total extensions, populated via tags, via split rules and via umbrella resolution, still empty, coverage, and instructions written. It also collects a missing-mnemonic log and fails the run under --strict, which is what npm run sync:check uses.

Since this was opened it also gained a dry-run mode in #171, and tests/catalog-coverage.test.mjs now asserts the invariant that matters most: a tag that resolves to nothing is a broken mapping and fails the test suite, rather than being reported as a warning somebody has to notice.

The branch conflicts with main now. If there is a specific warning case still going unreported, tell me which and I will add it.

Issue #98, opened 2026-05-15
feat: add interactive extension dependency graph and relationship mapping


ParthMozarkar:
### Description
Currently, the RISC-V Extension Landscape provides an excellent flat view of available extensions, but it is difficult for users to understand the complex inter-dependencies and bundling logic (e.g., how `G` bundles `I, M, A, F, D, Zicsr, Zifencei` or how `V` requires `D`).

### Proposed Solution
Introduce a comprehensive dependency management system and visualization layer:
1. **Dependency Engine**: A structured data mapping of requirements and bundling logic across the ISA.
2. **Interactive Visualization**: An SVG-based graph that allows users to see the "flow" of dependencies.
3. **Contextual Badges**: Tags within the extension detail panel that show what an extension requires and what other extensions depend on it.

### Impact
This improves the educational value of the tool, helping architects and developers understand the hardware requirements for specific extension combinations.

rpsene:
Closing this: two of the three parts landed, and the third was overtaken by a
different design.

**Dependency engine — done.** `src/isaGraph.js` carries the resolver and
`src/isa-dependency-graph.json` holds 219 nodes, seeded from riscv-unified-db by
`scripts/seed-dependency-graph.mjs` and checked against it in CI by
`npm run graph:check`, so the bundling and requirement logic you describe (`G`
pulling in its members, `V` requiring `D`) is modelled from the upstream source
rather than hand-maintained.

**Contextual behaviour — done, though not as badges.** Selecting an extension
resolves its requirements and auto-adds them, reporting what it did
("Auto-added: ... (Required dependency)"), and the details panel shows the real
floor a selection implies. That covers the "what does this require" half.

**Interactive SVG graph — not built, and not currently planned.** The need it
was aimed at got answered another way: the full-screen ISA builder studio (#244)
and the comparison dock and diff matrix (#223) are where relationships are
explored now.

Thanks for the write-up. If you still want the visual graph specifically, please
open a fresh issue scoped to just that, so it is tracked as the one open piece
rather than as a feature request that is mostly already shipped.

PR #99, opened 2026-05-15
feat: add interactive extension dependency graph and relationship visualization

ParthMozarkar:
### Overview
This PR introduces an interactive dependency graph and relationship mapping system to help users visualize and navigate the complex hierarchy of RISC-V extensions.

### Key Features
- **📊 Interactive SVG Graph**: A new modal-based visualization that maps extension relationships (Requires vs. Bundles) using a layered layout algorithm.
- **🔗 Navigation Badges**: Integrated relationship tags in the extension detail panel allowing for quick "jump" navigation between dependencies.
- **⚙️ Graph Logic Utilities**: New traversal logic (Topological sort, Transitive closure) for dependency analysis.
- **🛡️ Data Integrity Tests**: Added automated tests to ensure the dependency graph remains cycle-free and resolves to valid extension IDs.

### Technical Changes
- Created `src/components/DependencyGraph.jsx`, `DependencyModal.jsx`, and `DependencyBadges.jsx`.
- Implemented graph traversal algorithms in `src/utils/graph.js`.
- Added structured relationship data in `src/data/extension_dependencies.json`.
- Updated `riscv_extensions.json` with requirement metadata for core extensions.

### Verification Results
Ran the full test suite with 100% pass rate:
- **Total Tests**: 101 passed
- **New Coverage**: Unit tests for graph logic and integrity checks for the new dependency dataset.

Issue : #98


rpsene:
Keeping this one open rather than closing it with the others, because the idea is right and the ground has shifted in its favour.

Since it was opened, the dependency data landed in src/isa-dependency-graph.json: 228 nodes, one per catalogue extension, seeded from riscv-unified-db with a citation on every edge, plus src/isaGraph.js providing closure, resolveSelection and explain. Dependency resolution is live in the ISA Configuration Builder, so selecting D pulls in F and the tooltip names why.

What is still missing is exactly what this PR set out to build: a visualisation. So the gap is real.

Two things would need to change before it can be merged:

It ships its own src/data/extension_dependencies.json. That would be a second source of truth alongside the seeded graph, which is the drift trap we already have one parser for. It should read the existing graph instead.

It introduces jest alongside the existing node:test suite, which is 11 files and 123 tests run by npm test in CI. Two runners in one repo needs a deliberate decision, and my preference is to stay on node:test since it needs no dependencies.

It also conflicts with main now, and at 7119 insertions it is not reviewable in one pass regardless.

If you are still interested, a smaller PR that renders the existing graph, with no new data file and no new runner, would be very welcome and I will review it quickly. If you would rather not, say so and I will close this without any hard feelings, since the data layer landing was partly informed by your original proposal.

rpsene:
Closing this, and I want to give the real reason rather than a polite one.

The dependency data you proposed now exists. src/isa-dependency-graph.json holds 228 nodes, one per catalogue extension, seeded from riscv-unified-db with a citation on every edge, and src/isaGraph.js provides closure, resolveSelection and explain. Resolution is live in the ISA Configuration Builder: selecting D pulls in F, locked extensions show what requires them, and the tooltip names the reason. Your original proposal is part of why that got built.

What is still missing is the visualisation, which is what this PR set out to add. I am closing it anyway, because a visualisation presents information the tool already surfaces at the point where it matters, and that puts it behind other work rather than in front of it.

Being straight about the cost as well: 7119 insertions, it conflicts with main, it ships its own src/data/extension_dependencies.json which would be a second source of truth alongside the seeded graph, and it introduces jest alongside the existing node:test suite of 12 files and 125 tests. Any of those is surmountable. Together they are a larger review than the feature warrants right now.

If you want to come back to it, the version I would merge is small: render the existing isa-dependency-graph.json, no new data file, no second test runner. Say the word and I will review it quickly.

Sorry it sat unreviewed for three months. That is the part I would undo if I could.

PR #114, opened 2026-05-17
feat: Add automated tests for synchronization tooling

AmitShah13:
## Summary

Added lightweight automated tests for synchronization tooling using Node.js `assert`.

## Features

* Added synchronization tooling test script
* Added tests for instruction key normalization
* Added tests for extension alias reconciliation
* Added tests for dry-run flag detection
* Added encoding format validation tests
* Added match/mask hexadecimal consistency validation tests

## Why

The synchronization tooling previously had no automated test coverage.

This contribution improves reliability and regression safety for synchronization-related workflows including:

* instruction normalization
* alias reconciliation
* dry-run execution logic
* encoding validation
* match/mask consistency validation

## Benefits

* improves tooling reliability
* helps detect regressions
* improves contributor validation workflows
* supports safer future tooling refactors

## Testing

Tested locally by:

* running `node scripts/test_sync.mjs`
* verifying all assertions pass successfully

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "build": "webpack",
   "predeploy": "npm run build",
   "deploy": "gh-pages -d dist --branch gh-pages --dotfiles",
-  "test": "echo \"Error: no test specified\" && exit 1"
+  "test:sync": "node scripts/test_sync.mjs"
 },
   "keywords": [],
   "author": "",
--- a/scripts/test_sync.mjs
+++ b/scripts/test_sync.mjs
@@ -52,5 +52,41 @@ assert.strictEqual(
   isDryRun([]),
   false
 );
+function isValidHex(value) {
+  return /^0x[0-9a-fA-F]+$/.test(value);
+}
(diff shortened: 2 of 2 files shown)


rpsene:
Thanks. Closing as largely covered by what has since landed.

The suite is now 11 files and 123 tests, run by npm test in CI on every PR. For the sync specifically, tests/catalog-coverage.test.mjs asserts the invariants that catch real breakage: a tag never resolving to nothing, umbrellas resolving with members that exist, CSR coverage ratcheting, and no withdrawn instruction being published. tests/data-integrity.test.mjs covers encoding shape, match-within-mask and id uniqueness.

The sync script also self-checks: it asserts expected instruction counts for B, Zbb, Zk, Zkn and Zks, and refuses to write when validation fails. That guard has already caught three real changes this week.

The branch also introduces a parallel runner alongside node:test. If there is a specific sync behaviour still untested, I would rather add it to the existing suite.

PR #119, opened 2026-05-18
fix: Add missing deprecated instructions (SLO, SLOI, SRO, SROI) to the UI visualizer

pranjal2004838:
### Description

The B extension snapshot deprecated four instructions: `SLO`, `SLOI`, `SRO`, and `SROI`. While these exist in `src/riscv_extensions.json` with their complete encoding metadata, they were previously omitted from the instruction groupings in `src/risc_v_visualizer.jsx`. As a result, they were effectively invisible in the Instruction Set Snapshot and mnemonic navigation UI.

This PR addresses the omission by adding the missing mnemonics to the `B` and `Zbb` instruction groupings, successfully resolving the discrepancy between the underlying JSON catalog and the frontend rendering.

### Changes Made
- Added `SLO`, `SLOI`, `SRO`, and `SROI` to the `B` and `Zbb` instruction arrays within `src/risc_v_visualizer.jsx`.
- Added standard code comments indicating these are deprecated instructions to match surrounding conventions.

### Validation
- Validated via a local build check (`npm run build`).
- No build regression introduced; compiles cleanly.

### DIFF
--- a/src/risc_v_visualizer.jsx
+++ b/src/risc_v_visualizer.jsx
@@ -1510,6 +1510,9 @@ const RISCVExplorer = () => {
 
       // Zbb: Logical operations
       'ANDN', 'ORN', 'XNOR',
+      // Zbb: Deprecated logical operations
+      // adding these manually bcoz they are missing lol
+      'SLO', 'SLOI', 'SRO', 'SROI',
       // Zbb: Count leading/trailing zeros and population count
       'CLZ', 'CTZ', 'CPOP',
       'CLZW', 'CTZW', 'CPOPW',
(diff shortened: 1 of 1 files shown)


wmat (approved):
LGTM

rpsene:
Thanks for looking at this, but I need to go the other way, and I want to explain why rather than just decline.

SLO, SLOI, SRO and SROI were removed from the catalogue in #175 because they are not ratified. They come from the draft bitmanip proposal and were dropped before ratification. The evidence: upstream riscv-opcodes rv_zbb and rv64_zbb no longer list them, and with them present Zbb reported 28 instructions where the ratified set has exactly 24.

For a reference tool, publishing an encoding that does not exist is worse than omitting one, since somebody can implement against it. There is now a test pinning both the withdrawn mnemonics and the Zbb count.

If the goal is to show withdrawn instructions deliberately, as history, that is a reasonable feature, but it needs to be visibly labelled as unratified rather than listed alongside real ones.

PR #120, opened 2026-05-20
Bug: Fix SPA routing for Docker nginx deployments

AmitShah13:
## Summary

Added nginx configuration for SPA routing support in Docker deployments.

## Changes

* added `nginx.conf`
* configured SPA fallback routing using `try_files`
* updated `Dockerfile` to copy nginx configuration into container

## Why

Currently, Docker nginx deployments return `404` errors when:

* refreshing nested routes
* opening shared links directly
* navigating directly to SPA routes

This happens because nginx attempts to resolve routes as physical file paths instead of serving `index.html`.

The added configuration ensures unknown routes correctly fall back to the React SPA entrypoint.

## Testing

Verified configuration changes locally and confirmed nginx SPA fallback configuration is correctly included in the Docker image setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ FROM nginx:alpine
 # Copy the build output from Stage 1 to Nginx's html directory
 # Note: If you use CRA, change '/app/dist' to '/app/build'
 COPY --from=build /app/dist /usr/share/nginx/html
+COPY nginx.conf /etc/nginx/conf.d/default.conf
 
 EXPOSE 80
 
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,0 +1,13 @@
+server {
+    listen 80;
+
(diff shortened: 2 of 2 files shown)


rpsene:
Thanks for looking at this, but I think it fixes a problem the app does not have, so I am closing it.

SPA fallback routing exists so that a deep link like /extensions/Zbb reaches index.html instead of 404ing. That only matters with a client-side router. This app has no router: there is no routing dependency, no route definitions, and all state is in memory, so every URL is / and nginx already serves index.html for it.

The Dockerfile and docker-compose.yml already build and serve the app.

There is a related feature that would make this necessary, and it is a good one: #94 proposes shareable permalinks. If that lands using real paths rather than query parameters or a hash, the nginx try_files rule becomes genuinely required. Worth revisiting then, and I will link this PR from there.

PR #122, opened 2026-05-20
feat: add coverage_report.py to surface 137 unmapped extension gaps with classification

Rahulatram321:
## Problem
Instruction coverage gaps are currently hard to reason about because there is no structured classifier that separates extensions that are already mapped from extensions that are present under mismatched keys or absent entirely. This blocks maintainers from prioritizing fixes against the mentorship target of 137 unmapped extensions.

## Solution
This PR adds a reusable coverage analysis pipeline:
- Added `coverage_report.py` with typed classification logic and CLI output in Markdown/JSON.
- Classification statuses: `MAPPED`, `KEY_MISMATCH`, `ABSENT`.
- Added manual-mention annotation for ABSENT rows when a local `riscv-isa-manual` checkout is available.
- Added generated `COVERAGE.md` snapshot for immediate maintainer visibility.
- Added tests for all logic paths, CLI output modes, alias handling, and integration wiring.

Current snapshot numbers (`COVERAGE.md`):
- Total extensions: `220`
- MAPPED: `71`
- KEY_MISMATCH: `12`
- ABSENT: `137`

## Testing
```bash
pytest tests/test_coverage_report.py tests/test_extension_utils.py -v --cov=coverage_report --cov=extension_utils --cov-report=term-missing
```

```text
=============================== tests coverage ================================
_______________ coverage: platform win32, python 3.14.3-final-0 _______________

Name                 Stmts   Miss  Cover   Missing
--------------------------------------------------
coverage_report.py     142      0   100%
extension_utils.py      47      0   100%
--------------------------------------------------
TOTAL                  189      0   100%
```

## What this enables
Maintainers now get a deterministic gap report they can use to prioritize mapping work and to distinguish naming mismatches from truly absent extension data.

## Related
- Existing PRs: #91 (normalization baseline), #92, #93


rpsene:

PR #127, opened 2026-06-25
feat: add opcodes tag mapping to all 220 extension entries

### LINKED ISSUE #126: Add opcodes tag mapping to extension catalog for automated sync
## Problem

The sync script (`sync_instructions.mjs`) currently cannot automatically map instructions from `instr_dict.json` to entries in `riscv_extensions.json` because there is no machine-readable mapping between **riscv-opcodes** tag names (e.g., `rv_zba`, `rv64_zknd`) and catalog extension IDs (e.g., `Zba`, `Zknd`).

As a result, **104 of 114** opcode tags remain unlinked, preventing automatic synchronization of instruction data. This leaves **137 of 220** extensions without instruction encoding information.

## Examples of Mapping Issues

* **Base ISA**

  * `rv_i` → `RV32I` and `RV64I`
* **Width-specific variants**

  * `rv32_zknd` and `rv64_zknd` → `Zknd`
* **Compound tags**

  * `rv_d_zfa` → requires both `D` and `Zfa`
* **Missing catalog entries**

  * Some opcode tags have no corresponding extension in the catalog.

## Proposed Solution

Add a `tags` array to each extension entry in `riscv_extensions.json` that explicitly lists the corresponding **riscv-opcodes** tag(s).

### Example

```json
{
  "id": "Zba",
  "tags": ["rv_zba", "rv64_zba"]
}

{
  "id": "RV32I",
  "tags": ["rv_i"]
}

{
  "id": "Sstc",
  "tags": []
}
```

The synchronization script can then use this field to directly map opcode tags to extension IDs instead of relying on naming conventions or special-case logic.

ARYANPATEL-BIT:
# Closes #126

## Summary

This PR adds a `tags` array field to every extension entry in `riscv_extensions.json`. The new field provides a machine-readable mapping between **riscv-opcodes** tags and catalog extension IDs, enabling the sync script to automatically associate instructions from `instr_dict.json` with the correct catalog entries.

This is a non-breaking structural change that lays the groundwork for fully automated instruction synchronization in a follow-up PR.

## Changes

* Added a `tags` field to every extension in `riscv_extensions.json`.
* Extracted all **114** unique opcode tags from `instr_dict.json`.
* Mapped **100** opcode tags to their corresponding catalog extensions.
* Included support for:

  * Base ISA mappings (e.g., `rv_i`)
  * Width-specific variants (e.g., `rv64_zba`)
  * Compound tags (e.g., `rv_d_zfa`)

## Results

* ✅ **69** extensions now have one or more opcode tags.
* ✅ **151** extensions correctly have `tags: []` because they represent CSR-only, umbrella, capability, or supervisor extensions that do not directly map to opcode tags.
* ⚠️ **14** opcode tags remain unmapped because they correspond to:

  * New or draft extensions not yet present in the catalog.
  * The `rv_zicbo` split case, which requires additional handling.

## Why?

Without an explicit mapping, `sync_instructions.mjs` cannot reliably match opcode tags such as `rv_zba` or `rv64_zknd` to catalog IDs like `Zba` and `Zknd`. This prevents automated population of instruction metadata.

By introducing the `tags` field, the mapping becomes explicit, maintainable, and deterministic, eliminating the need for heuristic matching.

## Future Work

A follow-up PR will update `sync_instructions.mjs` to consume the new `tags` field and automatically populate instruction data from `instr_dict.json`.

### DIFF
--- a/src/riscv_extensions.json
+++ b/src/riscv_extensions.json
@@ -4,7 +4,9 @@
       "id": "RV32I",
       "name": "RV32I",
       "tags": [
-        "rv_i"
+        "rv_i",
+        "rv_zicsr",
+        "rv_zifencei"
       ],
       "desc": "Standard Integer Base (32-bit)",
       "use": "Microcontrollers, IoT",
@@ -628,7 +630,9 @@
       "name": "RV64I",
       "tags": [
         "rv64_i",
-        "rv_i"
+        "rv_i",
+        "rv_zicsr",
+        "rv_zifencei"
       ],
       "desc": "Standard Integer Base (64-bit)",
       "use": "Servers, Mobile, PC",
@@ -1408,7 +1412,9 @@
       "id": "RV32E",
       "name": "RV32E",
       "tags": [
-        "rv_i"
+        "rv_i",
+        "rv_zicsr",
+        "rv_zifencei"
       ],
       "desc": "Embedded Base (16 regs)",
       "use": "Tiny cores (Reduced silicon)",
@@ -2032,7 +2038,9 @@
       "name": "RV64E",
       "tags": [
         "rv64_i",
-        "rv_i"
+        "rv_i",
+        "rv_zicsr",
+        "rv_zifencei"
       ],
       "desc": "Embedded Base (64-bit, 16 regs)",
       "use": "Efficient 64-bit controllers",
@@ -2813,7 +2821,9 @@
       "name": "RV128I",
       "tags": [
         "rv64_i",
-        "rv_i"
+        "rv_i",
+        "rv_zicsr",
+        "rv_zifencei"
       ],
       "desc": "128-bit Address Space",
       "use": "Experimental/Research",
--- a/src/riscv_extensions.json
+++ b/src/riscv_extensions.json
@@ -3,6 +3,9 @@
     {
       "id": "RV32I",
       "name": "RV32I",
+      "tags": [
+        "rv_i"
+      ],
       "desc": "Standard Integer Base (32-bit)",
       "use": "Microcontrollers, IoT",
       "instructions": {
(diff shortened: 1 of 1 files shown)


caffeine-rohit (changes requested):
## one change needed from your end before merge:

**Base ISA tags missing `rv_zifencei` and `rv_zicsr`**

RV32I, RV64I, RV32E, RV64E, and RV128I tags need these two added.
The current JSX `extensionInstructions` explicitly groups FENCE.I and 
all 6 CSR instructions inside all base ISAs. When the sync script is 
rewritten in the next PR to route by tags instead of the JSX object, 
these instructions must be pulled into the base ISA entries automatically.
This is config that belongs in Step 1.

```
RV32I:  ["rv_i", "rv_zifencei", "rv_zicsr"]
RV64I:  ["rv64_i", "rv_i", "rv_zifencei", "rv_zicsr"]
RV32E:  ["rv_i", "rv_zifencei", "rv_zicsr"]
RV64E:  ["rv64_i", "rv_i", "rv_zifencei", "rv_zicsr"]
RV128I: ["rv_i", "rv_zifencei", "rv_zicsr"]

```
**I'll raise a follow-up PR to document and track the 14 currently unmapped tags as a dedicated repository artifact(scripts/catalog_gaps.json). Keeping that separate will let this PR stay focused on the mapping changes while ensuring the remaining gaps are tracked for the upcoming sync script rewrite.**

Everything else in this PR is correct. Good work on the mapping coverage aryan.

ARYANPATEL-BIT:
Good catch @caffeine-rohit . I've just pushed a commit adding rv_zifencei and rv_zicsr to all five base ISA entries (RV32I, RV64I, RV32E, RV64E, RV128I) to ensure FENCE.I and the CSR instructions correctly route to the base ISAs

PR #128, opened 2026-06-28
fix: remove duplicate bundle script from HTML template

sushilbhattarai45:
## Summary
This PR removes the manually added `/bundle.js` script tag from `public/index.html`.

The project already uses `HtmlWebpackPlugin`, which injects the generated bundle into `dist/index.html` during the Webpack build. Keeping the manual script caused the built HTML to include `bundle.js` twice.

## Changes
- Removed the hard-coded `/bundle.js` script tag from `public/index.html`.

## Results
- `dist/index.html` now includes only one generated bundle script.
- Avoids loading the application bundle twice.
- Keeps the HTML template compatible with Webpack’s automatic asset injection.

## Testing
- Ran `npm run build`.
- Confirmed the build completed successfully.
- Confirmed `dist/index.html` contains only one `bundle.js` script tag.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -7,6 +7,5 @@
 </head>
 <body>
   <div id="root"></div>
-  <script src="/bundle.js"></script>
 </body>
 </html>


caffeine-rohit:
Good catch @sushilbhattarai45 ! Removing this hardcoded tag is a great cleanup that prevents our React bundle from being downloaded and executed twice.

Code looks good. We can merge this safely.

sushilbhattarai45:
Thank you for looking into this pr.

rpsene:
Thanks for spotting this — closing as a duplicate of #75, which has just merged.

You and @ParthMozarkar independently produced **byte-identical** diffs for it. #75 was opened first (2026-05-08), so I merged that one. No reflection on this patch; it was the same correct fix.

For the record, the diagnosis was right and the bug was live in production. `webpack`'s `HtmlWebpackPlugin` already injects a script tag, so the hand-written one in the template meant the built page loaded the bundle twice — and the absolute `/bundle.js` path resolves off the project root, which 404s on a project Pages site:

```
https://rpsene.github.io/riscv-extensions-landscape/           200
https://rpsene.github.io/riscv-extensions-landscape/bundle.js  200   <- injected, correct
https://rpsene.github.io/bundle.js                             404   <- the hand-written tag
```

Every visitor was taking that 404. Now fixed on `main` and deployed.

Sorry it took this long to get to — please do send more.

PR #132, opened 2026-07-05
Add tags-based instruction sync foundation

Rahulatram321:
## Summary

- **Fixes the root architectural issue**: opcode tags from `instr_dict.json` (`rv_zba`, `rv_a`, etc.) now resolve to ISA manual catalog names (`Zba`, `A`, etc.) via `tags` arrays on each extension entry — eliminating the silent naming drift that caused ~104+ mismatches under a tag-join model.
- **Replaces JSX-driven sync membership** with tag-based matching in `sync_instructions.mjs`, using `scripts/lib/sync_rules.mjs` for prefix inference, split rules (`rv_zicbo` → Zicbom/Zicbop/Zicboz), and base-ISA overrides.
- **Adds baseline metrics** (`npm run metrics:baseline` → `reports/baseline_metrics.json`) so coverage is measurable on every PR.

### Baseline after this change

| Metric | Before (JSX sync) | After (tags sync) |
|--------|-------------------|-------------------|
| instr_dict coverage | ~72% (333 unmapped) | **97.5%** (30 unmapped) |
| Extensions with instructions | 71 | **83** |
| Unique synced mnemonics | 855 | **1,158** |

### New npm scripts

- `npm run bootstrap:tags` — seed/refresh `tags` on catalog entries
- `npm run sync:instructions` — tag-based merge from `instr_dict.json`
- `npm run metrics:baseline` — emit coverage report
- `npm run data:refresh` — runs all three in sequence

## Follow-ups (not in this PR)

- Remove `extensionInstructions` from JSX once UI reads mnemonic lists from catalog
- Data integrity validator + CI (weeks 7–8 plan)
- Encoder conflict graph + volume profile externalization
- Populate tags for 12 orphan opcode families (Zv*, Zbp, etc.)

## Test plan

- [x] `npm run data:refresh` completes without errors
- [x] `reports/baseline_metrics.json` shows 97.47% instr_dict coverage
- [x] RV32I contains no rv64-only instructions after tag overrides
- [ ] `npm run build` (requires `npm install` in CI/local)
- [ ] Spot-check Zba, Zicbom, Zabha, Zacas in the visualizer UI


caffeine-rohit:
Thanks for putting this together-- appreciate the effort on tag bootstrapping. 
That said, a lot of this overlaps with work already done/in-flight. 

Issues:

- Tag inference (inferTagsFromExtensionId) duplicates #127, which already hand-mapped all 114 tags from spec. Regex guessing breaks on umbrellas (Zk → rv_zk, which doesn't exist) and compound/width tags (rv_d_zfa, rv64_zba). The override table already proves inference alone isn't reliable.
- sync_instructions.mjs here skips umbrella resolution, RV32 shamt patching, and atomic file writes -- all of which #131 already handles properly.
- Regenerating riscv_extensions.json wholesale (43k additions) conflicts with #127's manual 406-line diff and makes this unreviewable line by line.
- npm test running baseline_metrics.mjs isn't right either -- tests should assert pass/fail, not emit a report that always exits 0.

Rather than carry two parallel implementations, let's converge on one foundation. #131 already has the harder problems solved (umbrella build, RV32 patch, atomic writes), makes more sense to land that first and then fix its remaining small bugs/edge cases than to introduce a second sync path here.

Happy to help scope a focused follow-up PR for tag bootstrapping if there's still a gap after #127/#131 land -- just narrower: fill missing tags only, no full regen, no inference.

rpsene:

PR #133, opened 2026-07-06
Externalize extension membership data

Rahulatram321:
## What changed
- moved extension instruction membership out of `src/risc_v_visualizer.jsx` into `src/extension_membership.json`
- moved Volume II membership into the same JSON data file
- updated `scripts/sync_instructions.mjs` to read membership data directly from JSON instead of scraping JSX
- cleared generated instruction maps before re-sync so removed memberships do not leave stale data behind
- updated docs to reflect the new source of truth

## Why
`extensionInstructions` was serving as canonical membership data and UI-local logic at the same time. That made the build-time aggregation depend on parsing JSX and kept Volume II membership hardcoded in the visualizer. Externalizing both data sets makes the sync path cleaner and gives the project a stable data source for future growth.

## Validation
- ran `node scripts/sync_instructions.mjs`
- ran `cmd /c npm install`
- ran `cmd /c npm run build`

## Notes
Build completed successfully. Current warnings are tooling-level:
- stale `baseline-browser-mapping` data
- stale `caniuse-lite` / Browserslist data
- webpack bundle size warnings


rpsene:
Superseded by #134 — I'll close this during tomorrow's review (2026-08-07) unless you'd rather keep it separate.

356301c is an ancestor of the #134 branch, and the three shared files are byte-identical between the two:

```
src/extension_membership.json    b8a7b083f149c52cfab5b79f96e16b9223213143
scripts/sync_instructions.mjs    04e430be4433f3ee9014ddd3bd50f6b2a3a47464
src/riscv_extensions.json        identical
```

So #134 is this commit plus the encoder validator work on top. Nothing here gets dropped — my review on #134 covers this content, and merging #134 lands it.

@Rahulatram321 if you'd prefer to land the membership externalization on its own first and rebase #134 on top, say so and I'll leave this open instead.

rpsene:
Closing — the commits here are contained in #134, so merging that lands this work too. Verified with `git merge-base --is-ancestor`; nothing is lost and your commit authorship stays intact. Context: #138

PR #134, opened 2026-07-07
Harden encoder validator checks

Rahulatram321:
## What changed
- extracted encoder validator logic into `src/encoder_validator.mjs`
- made invalid proposed input fail fast instead of continuing into conflict analysis
- added stricter validation for partial `Match`/`Mask` input
- added stricter validation for `match & ~mask != 0`
- added strict agreement checks between `Encoding` and `Match`/`Mask`
- skipped inconsistent existing instruction metadata and surfaced that as validator warnings
- added focused validator tests
- updated docs for the validator behavior

## Why
The validator was already useful, but it still allowed some invalid or inconsistent inputs to proceed into conflict reporting. That could produce confusing results and made the core decode logic harder to verify in isolation.

This PR improves correctness and maintainability by:
- separating the validator math from UI code
- making invalid inputs stop early
- making conflict classification easier to trust
- adding regression coverage for the core overlap logic

## Validation
- ran `npm test`
- ran `npm run build`

## Base
This PR is stacked on top of #133 and should initially target:
`codex/externalize-extension-membership`


rpsene (changes requested):
Reviewed against `origin/main` (b519ae1). Build compiles, all 7 tests pass.

The extraction looks right. I diffed top-level symbols between the old and new visualizer — the four that disappeared (`BIT_WIDTH`, `normalizeHexString`, `parseHexToBigInt`, `isSubsetPattern`) have zero remaining references there, so they're correctly internal to the module now, not dangling. The bit math in `patternsOverlap`, `isSubsetPattern` and `overlapExampleWord` checks out against the set semantics.

Two bugs, both in the validation this PR is named for.

**1. Over-wide hex is truncated instead of rejected.**

`normalizePatternSource` does `&= BIT_MASK_32` before validating anything:

```
match=0x33          mask=0x1FFFFFFFF   ->  match=0x33        mask=0xffffffff   errors: []
match=0xFFFFFFFFFF  mask=0xFFFFFFFFFF  ->  match=0xffffffff  mask=0xffffffff   errors: []
```

Paste a 40-bit value and you get a confident answer about a pattern you didn't type. #38 is open and fixes exactly this, so the two overlap — whichever lands second needs a rebase onto the new module.

**2. Invalid match/mask is silently dropped when an encoding is also supplied.**

```
encoding='-------------------------0110111', match='zzz', mask='qqq'
  ->  match=0x37  mask=0x7f   errors: []
```

Ordering problem in `normalizePatternSource`: both parse to `null`, the encoding branch fills them in, and the consistency check then compares the encoding against values derived from that same encoding — so it can never disagree. Without an encoding the same input does error, but with `"Unable to derive a complete Match/Mask pair from the provided input."` rather than anything about bad hex.

Minor:

- `encodingToMatchMask` is imported in `risc_v_visualizer.jsx` and never used.
- No test covers either bug above, or a negative `patternsOverlap` case. Both are one-liners.

Both fixes are local to `normalizePatternSource`: check width before masking, and check the hex parses before falling back to the encoding. The rest of the PR is good — and it's the first real test setup in this repo, which I want in.

Also: I'm closing #133, since 356301c is an ancestor of this branch and the shared files are identical. Reviewing this covers both.

rpsene:
Thanks for this, and I am sorry it sat since July.

Closing on reviewability rather than on merit. It is 3246 insertions against 2443 deletions across nine files including riscv_extensions.json, and it conflicts with main. That data file has been regenerated several times since, for CSR coverage in #174, withdrawn-instruction removal in #175 and the MOP expansion in #177, so this cannot be applied as it stands.

Hardening the validator is worth doing and some of it has moved since: #169 fixed the silent truncation of Match and Mask inputs wider than 32 bits, which previously reported a conflict against SC.W that the user never typed, and added Escape handling, a focus trap and dialog ARIA.

If you want to take another run at it, a smaller PR against current main covering one concern at a time would get reviewed quickly. Splitting the validator changes from the riscv_extensions.json changes would help most, since the data file moves under you.

PR #135, opened 2026-07-09
Feat: populate supervisor/CSR extensions 

ARYANPATEL-BIT:
Add structured CSR and behavioral data sourced from riscv-unified-db for the first batch of 12 supervisor/privilege-level extensions.

Extensions with CSR data (from riscv-unified-db csr/ directories):
- Smcntrpmf: mcyclecfg, mcyclecfgh, minstretcfg, minstretcfgh
- Smctr: mctrctl, sctrctl, vsctrctl
- Sscofpmf: scountovf
- Ssqosid: srmcfg
- Smepmp: mseccfg, mseccfgh
- Sstc: menvcfg, menvcfgh (root-level CSR files)
- Smstateen: mstateen0-3, hstateen0-3 + h-variants, sstateen0-3
- Sdtrig: tselect, tdata1-3, mcontext, scontext, hcontext, mscontext

Behavioral extensions (no instruction encodings or CSRs):
- Sv39: 39-bit VA paging, 3-level page tables
- Sv48: 48-bit VA paging, 4-level page tables
- Sv57: 57-bit VA paging, 5-level page tables
- Svnapot: NAPOT 64KB hugepage support via PTE N-bit encoding

Each entry now includes: long_name, type, state, ratification_date, and either a csrs map (address/priv_mode/length/desc per register) or a behavior field describing the hardware modification.

Data source: riscv-unified-db spec/std/isa/ext/ and csr/

### DIFF
--- a/src/riscv_extensions.json
+++ b/src/riscv_extensions.json
@@ -20953,8 +20953,7 @@
       "state": "ratified",
       "ratification_date": "2023-03",
       "requires": [
-        "S",
-        "SATP_MODE_BARE"
+        "S"
       ],
       "behavior": "The Bare translation mode disables virtual memory entirely. All virtual addresses equal physical addresses. Selected by writing satp.MODE = 0."
     },
@@ -21422,7 +21421,7 @@
       "state": "ratified",
       "ratification_date": "2024-06",
       "requires": [
-        "S,"
+        "S"
       ],
       "csrs": {
         "srmcfg": {
@@ -21553,7 +21552,7 @@
       "state": "ratified",
       "ratification_date": "2024-11",
       "requires": [
-        "S,"
+        "S"
       ],
       "csrs": {
         "mireg": {
@@ -22157,19 +22156,6 @@
       "type": "privileged",
       "state": "ratified",
       "ratification_date": "2023-03",
-      "requires": [
-        "REPORT_VA_IN_STVAL_ON_BREAKPOINT",
-        "REPORT_VA_IN_STVAL_ON_INSTRUCTION_ACCESS_FAULT",
-        "REPORT_VA_IN_STVAL_ON_LOAD_ACCESS_FAULT",
-        "REPORT_VA_IN_STVAL_ON_STORE_AMO_ACCESS_FAULT",
-        "REPORT_VA_IN_STVAL_ON_INSTRUCTION_PAGE_FAULT",
-        "REPORT_VA_IN_STVAL_ON_LOAD_PAGE_FAULT",
-        "REPORT_VA_IN_STVAL_ON_STORE_AMO_PAGE_FAULT",
-        "REPORT_VA_IN_STVAL_ON_INSTRUCTION_MISALIGNED",
-        "REPORT_VA_IN_STVAL_ON_LOAD_MISALIGNED",
-        "REPORT_VA_IN_STVAL_ON_STORE_AMO_MISALIGNED",
-        "REPORT_ENCODING_IN_STVAL_ON_ILLEGAL_INSTRUCTION"
-      ],
       "behavior": "Specifies that stval always writes the faulting virtual address on instruction/load/store/AMO address-misaligned or access-fault exceptions."
     },
     {
@@ -22183,9 +22169,6 @@
       "type": "privileged",
       "state": "ratified",
       "ratification_date": "2023-03",
-      "requires": [
-        "STVEC_MODE_DIRECT"
-      ],
       "behavior": "Specifies that the stvec register supports Direct mode (MODE=0), where all traps vector to the single base address."
     },
     {
@@ -22199,9 +22182,6 @@
       "type": "privileged",
       "state": "ratified",
       "ratification_date": "2023-03",
-      "requires": [
-        "STVEC_MODE_VECTORED"
-      ],
       "behavior": "Specifies that the stvec register supports Vectored mode (MODE=1), where interrupts vector to base + 4*cause."
     },
     {
@@ -22231,10 +22211,6 @@
       "type": "privileged",
       "state": "ratified",
       "ratification_date": "2024-10",
-      "requires": [
-        "TRAP_ON_UNIMPLEMENTED_INSTRUCTION",
-        "TRAP_ON_UNIMPLEMENTED_CSR"
-      ],
       "behavior": "Guarantees that no non-conforming extensions are present. All instructions not defined by ratified extensions raise illegal-instruction exceptions."
     },
     {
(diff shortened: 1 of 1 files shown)


src/riscv_extensions.json
```diff
@@ -21139,9 +21233,24 @@
...
+      "desc": "Quality of Service (QoS) is defined as the minimal end-to-end performance guaranteed in advance by a service level agreement (SLA) to a workload.",
       "use": "Per-thread QoS tagging",
-      "url": "https://github.com/riscv/riscv-isa-manual"
+      "url": "https://github.com/riscv/riscv-isa-manual",
+      "long_name": "Quality-of-Service Identifiers",
+      "type": "privileged",
+      "state": "ratified",
+      "ratification_date": "2024-06",
+      "requires": [
+        "S,"
```
caffeine-rohit: Minor typo mistake.

caffeine-rohit (approved):

src/riscv_extensions.json
ARYANPATEL-BIT: thanks for the review, I have fixed it in the latest commit and i would appreciate your review on the last 2 commit as well.

rpsene:
Flagging ahead of time: **I plan to close this PR during tomorrow's review (2026-08-07)**, in favour of the automated path in #137. Wanted to explain the reasoning and give you a chance to push back first.

## Why

**This PR is the output of the script in #137.** I cloned `riscv-unified-db` at `e909b6c` and ran that script against `origin/main`. It populates **the same 44 extensions** this PR touches, and **26 of its 39 rewritten `desc` values come out byte-identical to the ones here**. The two changesets are the same work — one done by hand, one generated.

Given that, it makes more sense to fix the generator once and re-run it than to hand-correct 44 entries here and then maintain both paths.

**There are two defects in this changeset that the generator fix also resolves:**

1. **23 of the 38 rewritten `desc` values are truncated or malformed.** `desc` is one of the few fields the UI actually renders (extension card, detail panel, search index), so this is a user-visible regression:
   - `Svbare`: `"Bare Mode"` → ``"This extension mandates that the `satp` mode Bare must"``
   - `Supm`: `"User Pointer Masking"` → `"Indicates that there is pointer-masking support available in user mode,"`
   - `Svukte`: `"User-Keyed TLB Entries"` → `"== \"Svukte\" Extension for Address-Independent Latency… Version 0.4"`

   The correct short label is already present in this PR as `long_name` (`"Bare virtual addressing"`, `"Pointer masking available in user mode"`), so the data is right — it just landed in the wrong field.

2. **Three `requires` entries don't resolve to any catalog extension**: `SXLEN` (Sv32, Sv39), `UXLEN` (Ssu32xl, Ssu64xl), and `Sm` (Svade, Svadu). `SXLEN`/`UXLEN` are implementation parameters of the same kind that `5562069` removed by hand — the generator's filter only catches all-caps names containing an underscore, so these slip through. `Sm` is a real UDB extension with no entry in our catalog, which is a coverage gap rather than a filter bug.

**The approval on this PR is also stale.** @caffeine-rohit approved on 2026-07-12, which was after `b1c5225` (batch 1, 12 extensions) but before `4288a47` (batch 2, 32 extensions) and `5562069`. So the approved portion is 12 of the 44 entries. 21 of the 23 malformed descriptions are in the never-reviewed batch 2. That isn't a criticism of anyone — without branch protection GitHub doesn't dismiss stale reviews, so the badge stayed green.

## What this work produced

To be clear that closing this isn't the same as discarding it — this PR is what surfaced everything above:

- It established that the UDB extraction approach works and is worth automating.
- Diffing it against the generator is what located both bugs in #137.
- It surfaced a genuine **upstream data bug** — `Svpbmt.yaml` in UDB carries Svbare's description verbatim. Filed as riscv/riscv-unified-db#2436.

## What happens next

Once the two fixes land on #137, the workflow regenerates this data correctly and opens a fresh PR. Net result is the same 44 extensions populated, minus the `desc` regression and the unresolved `requires`, and repeatable for every future UDB update instead of a one-off.

@ARYANPATEL-BIT if you'd rather keep this open — for instance if you want to land the corrected data here directly and use #137 only for future syncs — say so before tomorrow's review and I'll leave it open. Otherwise I'll close it then and we'll track the work on #137.

ARYANPATEL-BIT:
Thanks for the detailed write-up agreed, close it in favour of #137.

I went back through the data and your read is correct. The desc rewrites here are a regression, not an improvement: they replace a clean short label with a truncated fragment of the UDB description, and the correct label is already sitting in long_name. And SXLEN/UXLEN are the same class of implementation parameter I only removed by hand in 5562069 which means the filter, not the data, is the thing to fix.

So rather than hand-correct 44 entries here, I'll close this and move the fixes onto #137:

- **desc**: stop overwriting with the UDB description's first sentence. Keep the existing short label, fall back to long_name when there isn't one.
- **requires**: rather than extend the all-caps-with-underscore heuristic (which is what let SXLEN through in the first place), validate every extracted name against the catalog's own ids and drop anything that doesn't resolve, with a warning. That catches the next parameter without anyone having to spot it manually.

On Sm  agreed that's a coverage gap rather than a filter bug, so the validation above will drop it for now and log it. I'd rather add the catalog entry as a separate follow-up than widen #137 further; happy to open that this week.

Heads up on scope while you're reviewing #137: I'm also adding a `dependencies` array to every extension in riscv_extensions.json there, generated from the same requirements blocks. That's for the ISA Configuration Builder flat extension names, no version constraints, `[]` when there are none. Spec is confirmed with the contributor building the UI.

Thanks for filing the upstream Svpbmt/Svbare bug too. Once the two fixes land I'll run the workflow and let it regenerate this cleanly.

rpsene:
@ARYANPATEL-BIT Superseded by #137, which has now merged — closing this one, but the data you gathered here is all in `main`.

I checked rather than assumed. Comparing this branch against its merge-base (`b519ae1`) and then against `main`:

| | |
|---|---|
| extensions this PR populated | **44** |
| now covered on `main` | **44** |
| not covered | **0** |

So nothing is lost. Every extension you populated by hand has `long_name`, `type`, `state`, `ratification_date` and, where UDB has them, `csrs` or a `behavior` summary — produced by your own script in #137 rather than maintained by hand.

The one deliberate difference is `desc`. This PR rewrote **38** of them from UDB's `description`, and `main` keeps the curated originals in all 38 cases:

```
Sv32   here: "32-bit virtual address translation (3 level)"
       main: "Virtual Memory, 32-bit"
Sv39   here: "39-bit virtual address space paging scheme. Maps virtua…"
       main: "Virtual Memory, 39-bit VA"
```

That isn't a criticism of the wording — it's that `desc` is the short card label the UI renders (extension card, detail panel, search index), and UDB's `description` is a hard-wrapped block scalar that truncates into fragments. Your fix in #137 handles this correctly: `desc` now falls back to `long_name` only when an entry has none, and never overwrites a curated one. I verified against live UDB — 51 extensions updated, **zero** `desc` changes.

Closing this in favour of the automated path, as discussed in the #137 review. The weekly workflow now keeps this data current, so it should not need doing by hand again. Thanks for both.

PR #136, opened 2026-07-23
Add the Sh* augmented-hypervisor extensions (UDB completeness gap)

rpsene:
## Add the `Sh*` augmented-hypervisor extensions

Closes a completeness gap found by diffing this landscape against the **RISC-V Unified Database** (`riscv/riscv-unified-db`), the authoritative extension set.

### What was missing

The entire **`Sh*` "augmented hypervisor"** family exists in UDB but had **zero** representation here:

| Extension | Description |
|---|---|
| `Sha` | Augmented hypervisor (the umbrella bundle) |
| `Shcounterenw` | Hypervisor counter enable — writable `hcounteren` for all supported counters |
| `Shgatpa` | SvNNx4 guest translation for all supported S-mode modes, plus Bare |
| `Shtvala` | Hypervisor Trap Value (`htval`) provides all needed values |
| `Shvsatpa` | Virtual Supervisor address translation supports all supported S-mode modes |
| `Shvstvala` | Virtual Supervisor Trap Value (`vstval`) provides all needed values |
| `Shvstvecd` | Virtual Supervisor Trap Vector (`vstvec`) supports Direct mode |

These are privileged extensions mandated by the application-processor profiles (e.g. RVA23), so they matter for profile-completeness.

### Where they were added

All 7 are added to the **`s_trap`** category, alongside their supervisor analogues (`Sstvala`, `Sstvecd`, `Ssstateen`, …). Descriptions are taken from UDB's `long_name`. Two of them (`Shgatpa`, `Shvsatpa`) are address-translation-related and could alternatively live in `s_mem` — happy to move them if that fits the taxonomy better.

### Scope / diff

Pure addition — 7 entries, no other lines touched. The full UDB diff was **164 UDB extensions vs 218 here; 156 in common; these 8 in UDB but not in the landscape** (the 7 above + `Sm`).

### Known follow-up (not in this PR)

`Sm` (Machine mode) is also in UDB but appears here only as version-split `Sm1p11` / `Sm1p12` / `Sm1p13`. That's a naming/granularity difference rather than a clean omission, so it's left for a separate discussion on whether to add a canonical `Sm` entry.

### DIFF
--- a/src/riscv_extensions.json
+++ b/src/riscv_extensions.json
@@ -21510,6 +21510,62 @@
       "desc": "Trace & Instrumentation",
       "use": "Trace / instrumentation spec tag",
       "url": "https://github.com/riscv/riscv-isa-manual"
+    },
+    {
+      "id": "Sha",
+      "name": "Sha",
+      "tags": ["rv_sha"],
+      "desc": "Augmented hypervisor",
+      "use": "Bundle of hypervisor augmentation requirements (profiles)",
+      "url": "https://github.com/riscv/riscv-isa-manual"
+    },
+    {
+      "id": "Shcounterenw",
+      "name": "Shcounterenw",
+      "tags": ["rv_shcounterenw"],
+      "desc": "Hypervisor counter enable",
+      "use": "Writable hcounteren for all supported counters",
+      "url": "https://github.com/riscv/riscv-isa-manual"
+    },
+    {
+      "id": "Shgatpa",
+      "name": "Shgatpa",
+      "tags": ["rv_shgatpa"],
+      "desc": "SvNNx4 guest translation for all supported S-mode modes, plus Bare",
+      "use": "Guest-physical address translation requirement",
+      "url": "https://github.com/riscv/riscv-isa-manual"
+    },
+    {
+      "id": "Shtvala",
+      "name": "Shtvala",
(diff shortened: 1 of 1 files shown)


rpsene:
Superseded by #146.

Closing rather than updating: this branch dates from 2026-07-23, conflicts with `main` in `src/riscv_extensions.json`, and predates the dependency graph (#145), so a fresh branch was cleaner than a rebase.

The extension data here was sound — all seven verified against UDB `long_name` file by file, and `s_trap` is the right category. #146 carries it forward with three changes:

1. **`Sha` gets its graph node.** UDB gives it eight requirements (`H`, `Ssstateen`, and the other six `Sh*`), so as a bare entry it would have expanded to nothing. It now resolves to 12 extensions with a citation on every edge.
2. **`tags` are `[]` rather than `rv_sh*`.** Those tags map entries onto `riscv-opcodes` instruction groups; `rv_sha` and friends match zero entries in `instr_dict.json`, and 41 of 43 `s_trap` siblings use `[]`.
3. **`"instructions": {}`** added, matching every sibling in the category.

The follow-up note above about `Sm` is also resolved: UDB's `S` requires `Sm`, so #145 already carries it as a graph-only node (`catalogued: false`) to keep the graph closed. No canonical catalog entry needed.

PR #137, opened 2026-07-28
feat: automate syncing supervisor extensions from riscv-unified-db

ARYANPATEL-BIT:
## What this does

Automates pulling supervisor/CSR extension data from
[riscv-unified-db](https://github.com/riscv-software-src/riscv-unified-db)
into `src/riscv_extensions.json`, so we no longer have to check by hand
whether UDB has published data for the extensions we're waiting on.

## Changes

- **`scripts/sync_udb_extensions.cjs`** — dependency-free Node script that
  finds supervisor extensions still missing structured data, checks UDB for
  newly published YAML, and merges in CSR definitions, version/ratification
  info, dependencies, and descriptions. Idempotent — only rewrites the
  catalog when something actually changed.
- **`.github/workflows/sync-udb-extensions.yml`** — runs the script weekly
  (and on manual dispatch) and opens a review PR when new data is found.
  Runs are serialized with a `concurrency` group; third-party actions are
  pinned to commit SHAs.
- **`package.json`** — adds a `sync:udb` script for running it locally.

## Testing

- Verified locally against sample UDB fixtures: parses versions, CSRs,
  dependencies, and descriptions correctly.
- Confirmed idempotent — a second run with no new UDB data produces no diff.

## Before this can run automatically

- [ ] Enable **Settings → ActioActions to
      create and approve pull requests"** (needs admin).
- [ ] Merge to `main` — the schedule/manual triggers only activate on the
      default branch.
- [ ] Do one manual "Run workflow" after merge to confirm it works
      end-to-end against the live UDB repo.


rpsene (changes requested):
Two changes needed:

- Stop deriving desc from description — use long_name, or leave desc alone.
- Fix the param filter to /^[A-Z0-9_]+$/, and decide what to do about Sm.

Once those land, this is the right way to do what https://github.com/rpsene/riscv-extensions-landscape/pull/135 did manually — and I'd rather close https://github.com/rpsene/riscv-extensions-landscape/pull/135 and re-run the corrected script than maintain both.

rpsene (changes requested):
Thanks for this — the engineering around the workflow is genuinely careful, and I want to call that out before the findings below.

I reviewed this by cloning `riscv-unified-db` at `e909b6c` and running the script against `origin/main`, so everything here is reproducible:

```
node scripts/sync_udb_extensions.cjs ../riscv-unified-db
```

## What's good

- **All three actions are SHA-pinned, and I verified each against the GitHub API** — `actions/checkout@11d5960a`, `actions/setup-node@49933ea5`, `peter-evans/create-pull-request@c5a78066` all resolve to exactly the tags claimed. This is proper supply-chain hygiene.
- **Zero new npm dependencies** — `fs` and `path` only. No lockfile churn.
- **Idempotent, and I confirmed it.** Second run: gaps go 72 → 28, `Updated: 0`, file byte-identical. No reformat blast radius either — the 978 insertions are real content.
- `permissions:` correctly scoped to `contents: write` + `pull-requests: write`; `concurrency` with `cancel-in-progress: false`; output goes to a reviewable PR rather than a direct commit.
- `csrDefinedBy()` handles all three UDB `definedBy` shapes with indent-bounded scanning. The comment explaining why a bare regex would fail is exactly right.

## Blocking: this reproduces the `desc` regression from #135

Running the script populates **the same 44 extensions** as #135, and **26 of the 39 rewritten `desc` values are byte-identical to that PR's**. In my run, **29 of 39 rewritten descriptions were truncated or malformed**:

```
Svbare   before: "Bare Mode"
         after:  "This extension mandates that the `satp` mode Bare must"
Supm     before: "User Pointer Masking"
         after:  "Indicates that there is pointer-masking support available in user mode,"
Svukte   before: "User-Keyed TLB Entries"
         after:  "== \"Svukte\" Extension for Address-Independent Latency… Version 0"
```

This matters because `desc` is one of the few fields the UI actually renders — the extension card, the detail panel, and the search index all read it. The other fields this script adds (`long_name`, `type`, `state`, `ratification_date`, `requires`, `behavior`, `csrs`) are currently unrendered, so `desc` is the only user-visible effect, and right now it's a regression.

### Root cause — `scripts/sync_udb_extensions.cjs:334`

```js
const first = ext.description.split(/[.\n]/)[0].trim();
const candidate = truncateAtWord(first, 150);
if (candidate.length > (entry.desc || '').length) {
  entry.desc = candidate;
}
```

Two separate defects:

1. **Splitting on `\n`.** UDB descriptions are block scalars (`description: |`), hard-wrapped mid-sentence. Svbare's actual text is ``This extension mandates that the `satp` mode Bare must\nbe supported.`` — splitting on the newline drops "be supported." Because the truncation happens at the *line wrap*, `truncateAtWord`'s word-boundary logic never gets a chance to help.
2. **`candidate.length > entry.desc.length` treats longer as better.** For this field that's backwards — `desc` is a short card label, so the curated `"Bare Mode"` is better than a 53-character sentence fragment. This rule guarantees the fragment wins.

**Suggested fix:** don't derive `desc` from `description` at all. UDB's `long_name` is already the clean short label (`"Bare virtual addressing"`, `"Page-based memory types"`), and the script already extracts it correctly. Either leave `desc` untouched or seed it from `long_name`. Note that splitting on `.` alone wouldn't be enough — it still leaves the AsciiDoc-heading cases like `Svukte`.

## Blocking: inverted condition in the param filter

`scripts/sync_udb_extensions.cjs:92`

```js
if (name !== name.toUpperCase() || !name.includes('_')) {
  deps.push(name);
}
```

A name is filtered only if it's all-caps **and** contains an underscore, so all-caps names without one pass through:

| Name | Result |
|---|---|
| `SATP_MODE_BARE` | filtered ✅ |
| `TRAP_ON_UNIMPLEMENTED_CSR` | filtered ✅ |
| `SXLEN` | **admitted** ❌ |
| `UXLEN` | **admitted** ❌ |
| `MXLEN` | **admitted** ❌ |

My run leaked three unresolved `requires` targets: `SXLEN` (Sv32, Sv39), `UXLEN` (Ssu32xl, Ssu64xl), and `Sm` (Svade, Svadu). These are the same leaks that commit `5562069` on #135 patched by hand — which makes sense, since that PR was this script's output.

**Suggested fix:** filter on `/^[A-Z0-9_]+$/` (all-caps regardless of underscore). `Sm` is a different problem — it's a real UDB extension name that has no entry in our catalog, so it needs a coverage decision rather than a filter.

## Upstream bug worth reporting

`Svpbmt` and `Svbare` end up with identical descriptions, but that one is **not your bug** — `Svpbmt.yaml` in UDB genuinely contains Svbare's text:

```yaml
name: Svpbmt
long_name: Page-based memory types
description: |
  This extension mandates that the `satp` mode Bare must
  be supported.
```

I've filed this upstream: riscv/riscv-unified-db#2436.

Nothing here blocks this PR — the script copied the data faithfully. But it's a good argument for validating ingested data rather than trusting it, and until upstream fixes it, `Svpbmt` will keep importing the wrong text.

## Non-blocking

- **Pin or record the UDB commit.** The clone tracks `main` unpinned. Please include the synced UDB SHA in the generated PR body — I needed `e909b6c` to reproduce this review, and a future bad sync will need the same.
- **Silent failure on parse errors.** Every `parseYaml` failure is a `console.warn` + skip, and the script always exits 0. If UDB restructures its layout, the workflow would report "No new extensions found" instead of failing. Consider a non-zero exit when parse failures cross a threshold.
- **`supervisorPattern` includes `Sh`**, which overlaps #136 (Sh* augmented-hypervisor extensions). Not a conflict, just something to sequence.
- The workflow needs the org setting *"Allow GitHub Actions to create and approve pull requests"* enabled — thanks for flagging it; I'll handle that on my side.

## Summary

Two changes needed:

1. Stop deriving `desc` from `description` — use `long_name`, or leave `desc` alone.
2. Fix the param filter to `/^[A-Z0-9_]+$/`, and decide what to do about `Sm`.

Once those land, this is the right way to do what #135 did manually — and I'd rather close #135 and re-run the corrected script than maintain both.

rpsene:
@ARYANPATEL-BIT I've pushed to this branch rather than opening a competing PR, so your commits stay in the history. Here's exactly what changed and why — please push back on anything you disagree with.

**Both blocking items from my earlier review are fixed in your commits, and I confirmed it:** the param filter is now `/^[A-Z0-9_]+$/`, and `desc` only falls back to `long_name` when empty. Running the script against live UDB produces **zero `desc` changes**, so that regression is genuinely gone.

## What changed on my side

Between your last push and now, the repo gained a dependency graph (#145, #147): `src/isa-dependency-graph.json`, with a node for every catalog extension, a citation on every edge, and CI that fails if an extension has no node. That overlaps the dependency half of this PR, so I removed it and kept everything else.

Not a judgement on the code — it's that two parsers over the same UDB fields disagreed, and I could measure it. I ran your `extractDependencies` against all 169 UDB files:

**3 false negatives.** The line-based scan needs `name:` at the start of a line, so inline flow style is invisible:
```yaml
requirements:
  extension: { name: S, version: "~> 1.13" }   # returns [] silently
```
That's `Smcsrind`, `Ssqosid`, `Zcf`.

**9 false positives** from flattening `allOf`/`anyOf`/`oneOf`/`not` into one list — and two **invert** the meaning:

| flattened claim | what UDB says |
|---|---|
| `Zfinx` requires `F` | `not: F` — Zfinx *replaces* the F register file |
| `Zcmp` requires `Zcd` | `not: Zcd` |
| `C` requires `F` and `D` | C is legal with neither |

Your header comment did flag the flattening as a known simplification, so this isn't a hidden bug — it just turns out the distinction matters more than it looked. Worth saying: your insight about scanning **every** `requirements:` block regardless of nesting was right, and better than what I'd written. My seeder read only `requirements.extension` and missed `requirements.allOf[].extension` — about 44 blocks, including the whole `Zvl*b` chain. Reviewing this PR is what caught it, and #147 fixes it. Thank you for that.

## Concretely

- **Removed** `extractDependencies`, the dependency pass, and the `requires` write (~110 lines). The `requirements` block is now skipped outright rather than parsed and ignored.
- **Kept unchanged** everything else: CSRs, `long_name`, `type`, `state`, `ratification_date`, `behavior`. Verified against live UDB — 51 extensions updated, second run byte-identical.
- **Workflow**: clones `riscv/riscv-unified-db` (the `riscv-software-src/` path still redirects, but a redirect is one rename from a silent failure); records the UDB SHA in the generated PR body with repro commands — that was my outstanding non-blocking note, and reviewing this needed it by hand; reseeds the graph after the sync so a new extension arrives with its node already present; runs `npm test` before opening the PR; node 22 + `npm ci` to match `ci.yml`.
- **Resolved** the `package.json` conflict, keeping your `sync:udb` alongside main's test wiring.

Your SHA-pinned actions, the `concurrency` group, zero new dependencies, and the parse-failure threshold all stayed exactly as you wrote them — that engineering is why this is worth merging.

Two things still need @rpsene: enabling *"Allow GitHub Actions to create and approve pull requests"*, and a manual run after merge to confirm it end-to-end.

PR #168, opened 2026-08-19
feat(ui): premium UX polish, mobile responsiveness, and visual bug fixes

caffeine-rohit:
## Description

This PR introduces a comprehensive suite of UI polishing, accessibility improvements, and visual bug fixes aimed at elevating the application's user experience to modern, premium industry standards. We have overhauled several edge cases, refined keyboard navigation, and guaranteed perfect light/dark theme contrast—all while strictly preserving the zero-dependency, lightning-fast rendering speed of the core application. 

### 1. Mobile-Responsive Bottom Drawer
- Transformed the "Selected Details" sidebar into a sleek, swipe-up Bottom Drawer for mobile devices (`< 1024px`). 
- The drawer now pulls up smoothly from the bottom of the screen, maximizing grid space and working perfectly on phones, complete with a dedicated close icon.

Mobile Drawer View

### 2. Flawless Keyboard Navigation & Accessibility (a11y)
- **Focus Rings:** Implemented a global, consistent `:focus-visible` ring across all interactive elements (buttons, inputs, extension tiles). The glowing outline perfectly mimics the tile hover state, delivering a distinctly premium keyboard experience.
- **Interactive Tiles:** Fully re-engineered the Extension Tiles for keyboard users. You can now press `Tab` to seamlessly jump from tile to tile. Hitting `Enter` instantly opens the instruction details, and hitting `Space` adds/removes the extension from the Builder.
- **Screen Reader Support:** Added descriptive `aria-label`s to all icon-only controls (GitHub icon, clear search button, etc.) and restored native `title` attributes on disabled instruction pills to guarantee full screen-reader compliance.

### 3. Non-Intrusive Toast Notifications & Safe Timers
- Built a sleek, floating Toast notification system in the bottom-right corner. Notifications now slide in elegantly to provide clear visual feedback (e.g., "Copied instruction details!") without interrupting the user's flow or shifting layout elements.
- **Robust Timers:** Integrated `useRef` timer clearing across all copy actions. This eliminates race-condition bugs, ensuring that even if a user repeatedly clicks the copy button, the feedback icons and toast notifications will never glitch or overlap.
Toast Notifications

### 4. Zero-JS Pure CSS Tooltips
- Implemented a lightning-fast, pure CSS `[data-tooltip]` system to replace sluggish native browser titles. 
- These instant, beautifully styled tooltips are applied to the primary Builder Actions, Header buttons, and edge-aligned footer links, ensuring instantaneous feedback while keeping the UI completely uncluttered.

### 5. "Empty Search" State Illustration
- Introduced a polished "No Results Found" illustrative state. It clearly communicates an empty search instead of showing a broken, blank grid, and includes a convenient "Clear Search" quick-action button to instantly reset the view.
Empty Search State

### 6. Flawless Light Theme Contrast
- Fixed severe legibility issues in the detail card where the "Compressed Mapping" and "Compressed Equivalents" sections remained hardcoded to dark backgrounds (`slate-950/50`) in light mode. 
- These components now correctly adapt to the light theme's crisp surface color palette and typography, ensuring perfect legibility across both themes.

**Before**
Screenshot 2026-08-19 112226
Screenshot 2026-08-20 222351

**After**
Screenshot 2026-08-19 112648
Screenshot 2026-08-20 224732

### DIFF
--- a/src/ExtensionTile.jsx
+++ b/src/ExtensionTile.jsx
@@ -122,7 +122,7 @@ function ExtensionTile({
               transition: 'all 0.15s',
               padding: 0,
             }}
-            title={isLocked
+            data-tooltip={isLocked
               ? `Required by ${lockedBy.join(', ')} — remove dependent first`
               : (inWorkspace
                 ? `Remove ${data.id} from ISA Configuration Builder`
@@ -134,7 +134,7 @@ function ExtensionTile({
                   <path d="M1.5 4.5L3.5 6.5L7.5 2.5" stroke="currentColor" strokeWidth="1.75" strokeLinecap="round" strokeLinejoin="round" />
                 </svg>
               )
-              : <Plus size={9} />
+              : <Plus size={11} strokeWidth={2.5} />
             }
           </button>
         );
--- a/src/ExtensionTile.jsx
+++ b/src/ExtensionTile.jsx
@@ -51,7 +51,22 @@ function ExtensionTile({
   return (
     <div
       id={`ext-${data.id}`}
+      role="button"
+      tabIndex={dimmed ? -1 : 0}
       onClick={() => onSelect(data)}
+      onKeyDown={(e) => {
+        if (e.key === 'Enter') {
+          e.preventDefault();
+          onSelect(data);
+        } else if (e.key === ' ') {
+          e.preventDefault();
+          if (builderMode && !isDiscontinued) {
+            onToggleWorkspace(data.id);
+          } else {
+            onSelect(data);
+          }
+        }
+      }}
       className={[
         'ext-tile group relative rounded-lg border cursor-pointer select-none',
         isSelected ? 'ext-tile-active' : '',
@@ -97,16 +112,14 @@ function ExtensionTile({
         return (
           <button
             type="button"
+            tabIndex={-1}
             onClick={(e) => {
               e.stopPropagation();
               // The handler reports lock rejections itself.
               onToggleWorkspace(data.id);
             }}
+            className="workspace-tile-btn absolute top-1.5 right-1.5"
             style={{
-              position: 'absolute',
-              top: 6,
-              right: 6,
-              zIndex: 10,
               display: 'flex', alignItems: 'center', justifyContent: 'center',
               width: 18, height: 18,
               borderRadius: 5,
--- a/src/ExtensionTile.jsx
+++ b/src/ExtensionTile.jsx
@@ -97,14 +97,16 @@ function ExtensionTile({
         return (
           <button
             type="button"
-            data-in-workspace={inWorkspace ? 'true' : 'false'}
             onClick={(e) => {
               e.stopPropagation();
               // The handler reports lock rejections itself.
               onToggleWorkspace(data.id);
             }}
-            className="workspace-tile-btn absolute top-1.5 right-1.5"
             style={{
+              position: 'absolute',
+              top: 6,
+              right: 6,
+              zIndex: 10,
               display: 'flex', alignItems: 'center', justifyContent: 'center',
               width: 18, height: 18,
               borderRadius: 5,
@@ -122,7 +124,7 @@ function ExtensionTile({
               transition: 'all 0.15s',
               padding: 0,
             }}
-            data-tooltip={isLocked
+            title={isLocked
               ? `Required by ${lockedBy.join(', ')} — remove dependent first`
               : (inWorkspace
                 ? `Remove ${data.id} from ISA Configuration Builder`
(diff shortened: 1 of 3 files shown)


rpsene (changes requested):
Thanks, this is useful work. The mobile drawer and the empty search state are both real improvements and I verified them at 390px. A few things need to change before I merge.

## 1. The + button loses its class (src/ExtensionTile.jsx)

You replaced `className="workspace-tile-btn absolute top-1.5 right-1.5"` with inline styles. The positioning is equivalent, but five rules in index.css still target `.workspace-tile-btn` and are now dead:

* the `workspaceTileBtnIn` entrance animation
* `.ext-tile:hover .workspace-tile-btn { transform: scale(1.12) }`
* the two `prefers-reduced-motion` overrides
* the light theme `backdrop-filter: none`

Checked in the browser with the builder on. On main the button computes `animationName: workspaceTileBtnIn`, on this branch it computes `none`.

That animation is deliberate. It is the feedback that switching the builder on actually changed the page, which is what fixed the builder being undiscoverable in the first place. Please keep the class and layer your inline styles on top of it.

## 2. title -> data-tooltip drops accessible names

`data-tooltip` renders through a CSS `::after` on `:hover`. It is not an accessible name, and screen readers do not announce it.

I counted visible focusable controls with no accessible name: 0 on main, 3 on this branch.

* "Open the builder panel", icon only
* "View on GitHub", icon only
* the new mobile close button in the sidebar header, which has no label at all

Separately, hover-only tooltips never appear for keyboard users and cannot appear on touch, so the mobile users this PR is aimed at lose those hints completely.

Please add `aria-label` wherever `data-tooltip` replaced a `title` on an icon-only control, and add one to the new close button.

## 3. Tab and Enter on tiles does not work

The description says Tab moves focus between extension tiles and Enter adds one. The tiles are divs with no tabindex, so they cannot take focus. Calling `focus()` on a tile leaves `document.activeElement` unchanged. That also means `.ext-tile:focus-visible` can never match.

Either make the tiles focusable properly (tabIndex, a role, and an Enter/Space handler) or drop that rule and the claim from the description. The global `:focus-visible` ring does work on real buttons and inputs and is worth keeping either way.

## Smaller things

* `.toast-enter` and `.toast-exit` are defined in CSS but never applied in any JSX. Either wire them into the notice element or drop them.
* `showToast` starts a new 3500ms timer without clearing the previous one, so two toasts within 3.5s cut each other short. Hold the timer id in a ref and clear it before setting the next.
* Please put `if (!encoderValidatorResult?.proposed) return;` back in the copy handler. The button is disabled so it is not reachable today, but there is no reason to drop the guard.
* The instruction pill lost its title with nothing in its place. The `(no details yet)` text was the only thing explaining why a disabled pill is disabled.

## One note on the description

The blank grid in your "before" screenshot for the empty state is not what main does. A non-matching search leaves all 227 tiles visible and undimmed. The actual problem was that search looked like it did nothing, and your empty state does fix that, so keep it. I ran it against 11 queries including `addi`, `sc.w`, `0x1800202f` and `satp` and it never hides real matches.

Everything else looks fine. It merges cleanly onto main, build and tests pass, and all three commits are signed off.

rpsene (approved):
Verified all of it against a local merge, and every point is addressed. Nice work, and thank you for the thorough turnaround.

**The blocking one, the + button class.** \`workspace-tile-btn\` is back on the button, so the entrance animation, the hover scale and the reduced-motion overrides all work again. That animation matters more than it looks: it is the feedback that switching the builder on changed the page, which was the fix for the builder being undiscoverable in the first place.

**Accessible names.** Measured across 281 visible controls with the builder on and a tile selected: **0 with no accessible name**, down from 3. The 11 aria-labels cover the icon-only cases that lost theirs to `data-tooltip`, including the new mobile close button.

**Keyboard navigation, now real.** Tiles carry `role="button"` and take focus, so `.ext-tile:focus-visible` is reachable rather than dead CSS. `tabIndex={dimmed ? -1 : 0}` is a nice touch, keeping dimmed tiles out of the tab order.

One thing worth noting for anyone reading later, since it differs from the original description: Enter selects and **Space** adds to the builder. I verified Space, the builder badge went from empty to 1. That split is the better choice, because Enter matching click is the ARIA convention for `role="button"`, so no change needed.

**Smaller items, all done:** `.toast-enter` / `.toast-exit` removed rather than left as dead CSS, `showToast` clears the previous timer before setting the next so two toasts within 3.5s no longer cut each other short, the `encoderValidatorResult?.proposed` guard is back, and the instruction pill keeps its "(no details yet)" explanation.

Merges cleanly onto main, 123 tests pass, lint clean.

PR #222, opened 2026-08-26
feat: Implement expanded instruction details modal with full feature parity

caffeine-rohit:
## What this changes

Introduces a dedicated, full-screen overlay modal for instruction details that dynamically expands from the constrained sidebar view when requested.

**Key Implementations:**

- **Expanded Overlay Architecture:** Designed a responsive, full-screen modal utilizing the existing `riscv-card` glassmorphism tokens to ensure perfect aesthetic harmony.
- **100% Feature Parity & Context Density:** Restores all data from the small card, including encoding diagrams, variable chips, compressed mappings, and active profiles, while surfacing previously hidden database fields such as `compressedMapping.notes` and explicit RISC-V extension tags.
- **Continuous Internal Navigation:** Re-architected click handlers to prevent modal unmounting during equivalence traversal. Users can now continuously dive deep into `Standard Equivalent` or `Compressed Equivalent` instruction hierarchies without breaking out of the expanded context.
- **Edge-Aware Tooltip Positioning:** Added deterministic edge-alignment CSS utilities (`.tooltip-align-left`, `.tooltip-bottom-right`) to prevent tooltip clipping when rendering near viewport boundaries.
Screenshot 2026-08-26 225539

## Why

The existing right-side detail card provided insufficient horizontal space for deep instruction inspection, particularly for complex 32-bit encodings and multi-parameter compressed mappings.

This change provides users with a dedicated workspace for analyzing granular instruction data without cluttering the primary layout.

## How it was verified

- Validated UI integrity across both **Dark** and **Light** themes, ensuring all newly introduced interactive components strictly consume dynamic `var(--riscv-*)` tokens and `.riscv-btn` primitives.
- Verified deterministic tooltip positioning across varying desktop widths to guarantee zero viewport clipping.
- Confirmed stable React component lifecycle during rapid, recursive navigation inside the modal.
- `npm test` passes.
- `npm run build` succeeds.

## Sign-off

- Every commit is signed off (`git commit -s`), per the **DCO**.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -786,7 +786,8 @@ pre,
   border-color: var(--riscv-border);
 }
 
-:root[data-theme='light'] .text-cyan-300 {
+:root[data-theme='light'] .text-cyan-300,
+:root[data-theme='light'] .text-cyan-400 {
   color: #0f5a68 !important; /* Matches ext-tile cyan- text for consistency */
 }
 
@@ -803,8 +804,12 @@ pre,
   background: var(--riscv-surface-2) !important;
 }
 :root[data-theme='light'] .bg-slate-800,
-:root[data-theme='light'] .bg-slate-800\/70 {
+:root[data-theme='light'] .bg-slate-800\/70,
+:root[data-theme='light'] .bg-slate-800\/50 {
   background: var(--riscv-surface) !important;
+}
+:root[data-theme='light'] .bg-slate-600 {
+  background: var(--riscv-muted) !important;
 }
 :root[data-theme='light'] .border-slate-700 {
   border-color: var(--riscv-border-2) !important;
@@ -833,13 +838,15 @@ pre,
 :root[data-theme='light'] .text-yellow-300 {
   color: var(--riscv-warn) !important;
 }
+:root[data-theme='light'] .text-yellow-100,
 :root[data-theme='light'] .text-yellow-200 {
   color: #8a4a08 !important;
 }
 :root[data-theme='light'] .border-yellow-400 {
   border-color: #f2d4b6 !important;
 }
-:root[data-theme='light'] .bg-yellow-500\/10 {
+:root[data-theme='light'] .bg-yellow-500\/10,
+:root[data-theme='light'] .bg-yellow-900\/20 {
   background: #fdefe1 !important;
 }
 
@@ -847,9 +854,11 @@ pre,
   color: #8c2230 !important;
 }
 :root[data-theme='light'] .border-red-400,
-:root[data-theme='light'] .border-red-500\/60 {
+:root[data-theme='light'] .border-red-500\/60,
+:root[data-theme='light'] .border-red-600\/60 {
   border-color: #f3c3c3 !important;
 }
+:root[data-theme='light'] .bg-red-950\/40,
 :root[data-theme='light'] .bg-red-500\/10,
 :root[data-theme='light'] .bg-red-500\/5 {
   background: #fdeaea !important;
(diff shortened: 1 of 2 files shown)


rpsene (changes requested):
Thanks for this, the modal itself is well structured: Escape closes it, the backdrop click works, `stopPropagation` is in the right place, and `aria-modal` / `aria-labelledby` are set.

Three things need fixing before this can go in.

## 1. Light theme: unreadable text (blocker)

`index.css` remaps Tailwind's dark palette for light theme (see the block near "baked into the markup in ~33 places"). This PR introduces pairs where only one half of the pair is remapped, so foreground and background end up on the same side of the scale:

| Element | Foreground | Background | Contrast (light) |
|---|---|---|---|
| Match / Mask box when the search matches | `text-yellow-100` (not remapped) | `bg-yellow-500/10`, remapped to `#fdefe1` | **1.05:1, invisible** |
| "Discontinued" badge | `text-red-200`, remapped to `#8c2230` | `bg-red-950/40` (not remapped) | 3.47:1, fails AA |
| "Equivalent Instruction" link | `text-cyan-400` (not remapped) | white | 1.81:1, fails AA |

The cyan one is the easiest to see: `text-cyan-300` is already remapped to `#0f5a68` (7.83:1) for exactly this reason, so `cyan-400` just needs the same treatment.

The yellow highlight is the blocker. At 1.05:1 the value is not dimmed, it is gone.

Either add light remaps for the new classes alongside the existing ones, or use the `--riscv-*` tokens directly, which is the more durable option.

## 2. Light theme: dark plates

These are not remapped either and will paint as dark rectangles on a light page: `bg-slate-600`, `bg-slate-800/50`, `bg-yellow-900/20`, plus the hardcoded `#22d3ee` on the "Compressed Mapping" heading. (`bg-slate-800/70` and `border-slate-700` are already covered, so those are fine.)

## 3. Focus management

The dialog sets `aria-modal="true"` but focus is never moved into it, Tab is not held inside it, and focus is not restored to the Expand button on close. With the backdrop up, Tab walks into the page behind it.

`src/focusTrap.js` exists and `CompareView.jsx` already uses it (`focusableWithin` / `nextFocus`), so this should be able to follow that pattern closely.

## Note on #223

#223 contains this PR's commit (`7f04c600`) while also targeting `main`. Whichever lands first makes the other's copy redundant. Could you restack #223 on top of this one, so the two can merge in order?

rpsene (approved):
All three items check out. Verified against the diff rather than the summary, and CI is green.

| Item | Verified |
|---|---|
| `text-cyan-400` | Folded into the existing `text-cyan-300` rule — 7.83:1 |
| `text-yellow-100` / `text-yellow-200` → `#8a4a08` | 6.07:1 on `#fdefe1` |
| `bg-yellow-900/20`, `bg-slate-800/50`, `bg-slate-600` | All remapped |
| Discontinued badge → `--riscv-report-*` | 4.54:1 composited |
| "Compressed Mapping" heading | Hardcoded `#22d3ee` replaced by the class, so it inherits the remap |
| Focus management | Correct — detail below |

The focus trap is the part I looked at hardest, and it is right: guarded on `instructionExpandOpen`, captures `activeElement` before moving focus, focuses a `tabIndex={-1}` container, and only calls `preventDefault()` when `nextFocus` returns non-null — so a Tab that lands mid-range still falls through to the browser, which is the easiest thing to get wrong here. Cleanup restores focus to the Expand button. It follows `CompareView` closely.

Two corrections to the numbers, neither blocking:

**The yellow remap is 6.07:1, not 7.5:1.** It clears AA comfortably, so the fix stands, but it does not reach AAA. Worth correcting so nobody later cites 7.5 as headroom that isn't there.

**The Discontinued badge clears AA by 0.04.** `--riscv-report-tint` composites to `#f8e7ea` on the white card, putting `#c2334d` at exactly 4.54:1 against a 4.5 floor. It only holds because `.riscv-card` is `#ffffff` in light theme — the same badge on `--riscv-bg` (`#f7f5fb`) measures 4.22:1 and fails.

Merging this. Four spots still show the same half-a-pair pattern this review was about, plus one redundancy; I will push those as a follow-up rather than hold the PR:

- `border-slate-700/50` renders `#99a0aa`, a mid-grey border on a now-white plate. `border-slate-700` is remapped, but the `/50` variant is a separate class.
- `border-yellow-700/30` renders `#e1c5a0` on the cream plate.
- `decoration-cyan-400/30` and `decoration-cyan-300` stay bright while the link text they underline is now `#0f5a68`.
- Two effects keyed on `instructionExpandOpen` both close on Escape. Idempotent, so harmless, but one should lose its Escape branch.

None of those break readability — they read as heavier or lighter than intended, not invisible.

Thanks for the thorough turnaround on this, and for restacking #223.

PR #227, opened 2026-08-26
docs: add AGENTS.md for AI agent contributors

ARYANPATEL-BIT:
Adds a root-level `CLAUDE.md` so AI coding agents (and new human
contributors) start with this repo's real invariants instead of guessing
at them.

### Why

Several things here are expensive to get wrong and aren't obvious from
reading the code:

- `src/instr_dict.json` is hand-maintained, not regenerated a
  "helpful" regenerate silently drops the entries upstream lacks
- every dependency-graph edge needs a citation or `validateGraph` fails
- the classic JSX transform means the unused-looking `React` import in
  every `.jsx` must stay
- every commit needs a DCO sign-off

An agent without this context opens PRs that fail CI, or worse, quietly
delete data.

### Contents

- Setup and a full command table, including how to run a single test
- Architecture: entry point, module map, data flow, which modules stay pure
- Recipes for adding an extension, an instruction, or a profile
- Conventions observed in the codebase (Prettier config, classic JSX
  transform, citation requirement, `node:test` style)
- Gotchas / do-not-touch list
- Workflow: branch prefixes, DCO, pre-PR checks, what CI enforces

Links out to `README.md` and `CONTRIBUTING.md` rather than duplicating
them, and is kept short deliberately the file is loaded into context on
every turn, so length costs tokens.

### Notes

Docs-only; no code or data changes. Happy to rename to a tool-neutral
`AGENTS.md`, move it under `docs/`, or trim anything that reads as
redundant with `CONTRIBUTING.md`.


rpsene (approved):
Verified every claim in the file against the repo — all hold. Renamed to AGENTS.md with CLAUDE.md as a pointer, matching riscv-unified-db.

PR #273, opened 2026-09-03
fix(export): fold shorthand bundles in riscv-config format and exclude conflicting base letters

### LINKED ISSUE #271: fix(export): fold shorthand bundles in riscv-config format and exclude conflicting base letters
### Description & Root Cause

1. **Shorthand bundle subsets emitted in `riscv-config` export format**:
In `src/exportUtils.js`, the `riscv-config` export format formats the ISA string and configuration file to meet the schema requirements of RISC-V International's `riscv-config` validator (`riscv/riscv-config`).
The code comment in `src/exportUtils.js` specifically notes:
```javascript
// V is a shorthand for Zve64d plus a 128-bit VLEN floor, so riscv-config
// treats listing both as an error outright: "V and Zve* cannot exist
// together". clang tolerates the redundant form, which is why -march keeps
// it, but this format must not. Same rule as Zkn absorbing Zbkb.
```
While `V` folds its vector members (`^Zv(e|l)`), the shorthand bundles defined in `marchUtils.js` (`SHORTHAND_BUNDLES`: `Zkn`, `Zks`, `Zk`) were not folded in `rcCandidates`. As a result, exporting a selection with `Zkn` outputs:
```yaml
  ISA: RV64IZbkb_Zbkc_Zbkx_Zkn_Zknd_Zkne_Zknh
```
When validated with `riscv-config -ispec <file>.yaml`, the official validator rejects the ISA string outright:
> *"In presence of Zkn the subsets must be ignored in the ISA string."*

2. **Base integer letters (`I` and `E`) permitted as single-letter extensions**:
In `buildIsaConfigYaml`, single-letter extensions are filtered with:
```javascript
const filteredSingles = singleLetters.filter(l => l !== baseInfo.base.toUpperCase());
```
Because this only filters the chosen base's own letter, passing or selecting `I` alongside `RV32E`/`RV64E` leaves `I` in `filteredSingles`. This produces an invalid ISA string `RV32EI`, emits `I` into `extensions:`, and produces `RV32EI` in the `riscv-config` ISA string without warnings. Under RISC-V unprivileged specification §27, `I` and `E` define mutually exclusive base integer architectures and cannot appear as extension letters.

---

### Expected Behavior

1. In the `riscv-config` format, whenever a shorthand bundle (`Zkn`, `Zks`, `Zk`) is present in `selectedIds`, its member subsets (`Zbkb`, `Zbkc`, `Zbkx`, `Zknd`, `Zkne`, `Zknh`, `Zksed`, `Zksh`, `Zkr`, `Zkt`) should be absorbed and omitted from the ISA string, with a comment explaining that they were folded into the shorthand.
2. In all export formats, `I` and `E` should not be admitted as single-letter extensions, and selecting a base letter mutually exclusive with the chosen base ISA should be excluded and reported in `warnings`.

---

### Proposed Solution

1. Import `SHORTHAND_BUNDLES` in `src/exportUtils.js`.
2. In the `riscv-config` format, filter out any candidate sub-extension covered by an active shorthand in `selectedIds`, and emit a comment block detailing the folded sub-extensions.
3. In `buildIsaConfigYaml`, guard single-letter partitioning against `I` and `E`, emitting an explicit warning if an incompatible base letter is dropped.
4. Add regression tests in `tests/export.test.mjs`.

tagadearpit:
Closes #271.

Supersedes #272 with verified DCO sign-off matching the commit author email (`143588157+tagadearpit@users.noreply.github.com`).

### Summary of Changes

1. **Absorb shorthand bundles in `riscv-config` export format**:
   - In `src/exportUtils.js`, the `riscv-config` export comments explicitly note:
     ```javascript
     // V is a shorthand for Zve64d plus a 128-bit VLEN floor, so riscv-config
     // treats listing both as an error outright: "V and Zve* cannot exist
     // together". clang tolerates the redundant form, which is why -march keeps
     // it, but this format must not. Same rule as Zkn absorbing Zbkb.
     ```
   - While `V` absorbed `Zv(e|l)`, shorthand crypto bundles (`Zkn`, `Zks`, `Zk`) were not folded in `rcCandidates`. As a result, selecting `Zkn` emitted both the shorthand and its component subsets (`Zbkb`, `Zbkc`, `Zbkx`, etc.) in the `ISA:` string, which `riscv-config` rejects with *"In presence of Zkn the subsets must be ignored in the ISA string."*
   - Imported `SHORTHAND_BUNDLES` from `./marchUtils.js` and filtered out covered member subsets in `rcCandidates` when their shorthand is selected.
   - Emitted a comment block documenting the folded shorthand sub-extensions, matching the existing `V` folding pattern.

2. **Exclude mutually exclusive base letters from extensions in configuration exports**:
   - In `buildIsaConfigYaml`, single-letter extension filtering previously only excluded the chosen base's own letter (`l !== baseInfo.base.toUpperCase()`). If `I` was selected with `RV32E`, `I` remained in `filteredSingles`, producing `RV32EI` in `isa_string`, `rcIsa`, and listing `I` under `extensions:`.
   - Under RISC-V Unprivileged Specification §27, `I` and `E` specify mutually exclusive base integer architectures and cannot appear as extension letters.
   - Handled `I` and `E` explicitly during single-letter partitioning, ensuring neither is admitted as an extension letter and emitting an explicit warning if an incompatible base letter was selected.

3. **Tests**:
   - Added unit test in `tests/export.test.mjs` verifying that `Zkn` absorbs its member subsets (`Zbkb`, `Zbkc`, `Zbkx`, `Zknd`, `Zkne`, `Zknh`) in `riscv-config` format and notes the omission in comments.
   - Added unit test in `tests/export.test.mjs` verifying that mutually exclusive base letters (`I` and `E`) are not emitted as extensions and that appropriate warnings are generated.

### DIFF
--- a/src/exportUtils.js
+++ b/src/exportUtils.js
@@ -8,13 +8,13 @@
  *
  * COMPILER COMPATIBILITY SCOPE (per extension family):
  *   Scalar crypto (Zk, Zkn, Zks, Zbkb, etc.): supported since ~GCC 12-13 / LLVM 14-15.
- *   Vector crypto (Zvkned, Zvbb, Zvbc, Zvkg family): GCC 14+ / LLVM 18+ (stable).
+ *   Vector crypto (Zvkned, Zvbb, Zvbc family): requires GCC 14+ / non-experimental LLVM 18+.
  *   Zve/Zvl sub-profile tokens: exact min version unconfirmed; verify with your toolchain.
  *   Base/gc extensions: universally stable.
  *   Full details and CI gap: see marchUtils.js COMPILER VERIFICATION SCOPE.
  */
 
-import { buildMarchString, BASE_ISA_IDS, BASE_ISA_PREFIX_MAP } from './marchUtils.js';
+import { buildMarchString, BASE_ISA_IDS, BASE_ISA_PREFIX_MAP, SHORTHAND_BUNDLES } from './marchUtils.js';
 import { buildCombinedCatalog } from './marchUtils.js';
 import { resolveParams } from './isaGraph.js';
 import { DEPENDENCY_GRAPH } from './isaGraph.js';
@@ -161,6 +161,14 @@ export function buildIsaConfigYaml(selectedIds, allExts, options = {}) {
     }
 
     if (id.length === 1) {
+      if (idUpper === 'I' || idUpper === 'E') {
+        if (idUpper !== baseInfo.base.toUpperCase()) {
+          warnings.push(
+            `"${id}" was dropped: it names a base ISA that is mutually exclusive with ${baseInfo.id}.`,
+          );
+        }
+        continue;
+      }
       singleLetters.push(idUpper);
       allExtTokens.push(idUpper);
     } else {
@@ -374,7 +382,17 @@ export function buildIsaConfigYaml(selectedIds, allExts, options = {}) {
       .sort((a, b) => RC_LETTER_ORDER.indexOf(a) - RC_LETTER_ORDER.indexOf(b))
       .join('');
     const vPresent = rcSingles.includes('V');
-    const rcCandidates = vPresent ? zExts.filter((z) => !/^Zv(e|l)/i.test(z)) : zExts;
+    const shorthandAbsorbed = new Map();
+    for (const [shorthand, members] of Object.entries(SHORTHAND_BUNDLES)) {
+      if (selectedIds.includes(shorthand)) {
+        for (const member of members) shorthandAbsorbed.set(member, shorthand);
+      }
+    }
+    const rcCandidates = zExts.filter((z) => {
+      if (vPresent && /^Zv(e|l)/i.test(z)) return false;
+      if (shorthandAbsorbed.has(z)) return false;
+      return true;
+    });
     const rcKnown = rcCandidates.filter((z) => RISCV_CONFIG_SUB_EXTENSIONS.has(z)).sort(riscvConfigSort);
     const rcDropped = rcCandidates.filter((z) => !RISCV_CONFIG_SUB_EXTENSIONS.has(z)).sort();
     const rcAbsorbed = vPresent ? zExts.filter((z) => /^Zv(e|l)/i.test(z)).sort() : [];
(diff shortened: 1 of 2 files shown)


rpsene (approved):
Thanks for the re-spin with the corrected sign-off, and for splitting this out of #272 cleanly.

I verified both fixes by running the same probe against `main` and against this branch rather than reading the diff alone. Both bugs are real and both fixes are correct.

**Shorthand bundles**

| Selection | `main` | this PR |
|---|---|---|
| `Zkn` + its members | `RV64IZbkb_Zbkc_Zbkx_Zkn_Zknd_Zkne_Zknh` | `RV64IZkn` |
| `Zk` + `Zkn` | `RV64IZk_Zkn` | `RV64IZk` |
| `Zk` + `Zkn` + `Zks` | `RV64IZk_Zkn_Zks` | `RV64IZk_Zks` |

That third row is the one I expected to get wrong, and it is right: `Zk` absorbs `Zkn` but correctly does not absorb `Zks`, since `Zk` is the NIST suite (`Zkn` + `Zkr` + `Zkt`) while `Zks` is ShangMi. I also confirmed `Zk`, `Zkn` and `Zks` are all present in `RISCV_CONFIG_SUB_EXTENSIONS`, so folding the members never silently drops crypto from the string.

**Base letters**

`RV32E` + `I` gave `RV32EIM` before and now gives `RV32EM` with a warning; `RV64I` + `E` gave `RV64IEM` and now gives `RV64IM` with a warning. I checked the redundant paths (`RV64I` + `I`, `RV32E` + `E`) specifically, because the new `continue` skips `allExtTokens` where the old code filtered later, and the output is identical before and after. No collateral change.

Test suite is green here: 254 tests, 252 pass, 0 fail, 2 skipped for a missing UDB checkout.

Approving. Two small things I would like addressed, neither of which affects correctness.

**1. The doc-comment hunk contradicts its own cited source**

This is the only change unrelated to #271:

```diff
- *   Vector crypto (Zvkned, Zvbb, Zvbc, Zvkg family): GCC 14+ / LLVM 18+ (stable).
+ *   Vector crypto (Zvkned, Zvbb, Zvbc family): requires GCC 14+ / non-experimental LLVM 18+.
```

The intent is reasonable, since it makes the header agree with the string the file emits at `src/exportUtils.js:452`. But the header's following line points at `marchUtils.js` as the authority, and `src/marchUtils.js:29` lists the full family: `Zvkned, Zvbb, Zvbc, Zvkg, Zvksh, Zvksed`. So this trades one inconsistency for another, and resolves it in the direction that loses information: `Zvkg` is real and ratified, which makes the emitted line at 452 the incomplete one rather than the header.

Could you instead add the missing members to line 452 and leave line 11 naming the whole family, so all three agree? If you would rather keep it out of this PR entirely, dropping the hunk is also fine by me and I will open a separate issue.

**2. The test name over-claims**

`shorthand bundles (Zkn, Zks, Zk) absorb their member subsets` only exercises `Zkn`. Either rename it to say `Zkn`, or loop the three. I would prefer the loop, plus a case for `Zk` + `Zkn` together, because that case's correctness depends on `Object.entries` iteration order over `SHORTHAND_BUNDLES` and it will regress silently if anyone reorders that object.

Minor while you are in there: `assert.ok(isa.includes('Zkn'))` is also satisfied by `Zknd` on its own, so the positive assertion is weaker than it reads. Splitting `isa` on `_` and comparing tokens would be sturdier than substring matching. The negative assertions in the loop do cover the real case, so this is a nit rather than a defect.

**One observation, no action needed here**

The absorption loop duplicates what `src/marchUtils.js:455-458` already does. Since `exportUtils.js` now imports `SHORTHAND_BUNDLES` anyway, exporting a small `absorbedByShorthand(selectedIds)` helper from `marchUtils.js` would give both call sites a single implementation. Better as a follow-up than as growth on this PR, so I will file it separately.

PR #277, opened 2026-09-04
fix(march): handle extension version suffixes in -march string parser

### LINKED ISSUE #276: parseMarchString: extension version suffixes (e.g. rv64i2p0, rv32i2p1_m2p0, zba1p0) misparsed or dropped
### Description

In `src/marchUtils.js`, `parseMarchString` decodes `-march` strings into catalog extension IDs (e.g., when pasting an `-march` string into the Workspace / ISA Configuration Builder).

According to the **RISC-V Unprivileged ISA Specification Volume I, Chapter 27 ("ISA Extension Naming Conventions"), §27.2**:
> *"An extension name may be optionally followed by a version number of the form `major_number` or `major_number 'p' minor_number` where `major_number` and `minor_number` are integers (e.g., `rv64i2p0`, `rv32i2p1_m2p0`, `rv64i_zba1p0`, `zbb1p0`)."*

Compilers such as GCC and LLVM/Clang regularly emit and accept `-march` strings with explicit extension version numbers.

### Current Behavior

`parseMarchString` currently handles neither single-letter extension version numbers nor multi-letter version suffixes:

1. **Single-letter extensions with versions**:
   `parts[0]` is iterated character-by-character (`for (const ch of parts[0] || '')`). When given a string such as `rv64i2p0`:
   - `'i'` resolves to `RV64I`.
   - `'2'` and `'0'` are added to `unknownTokens`.
   - `'p'` (the delimiter between major and minor version digits) is looked up in the catalog, matching the retired `P` extension, and triggers a false warning:
     `"P" is in the extension catalog but is NOT a valid -march token (UI grouping tag or non-ISA entry). It has been ignored.`

2. **Multi-letter extensions with versions**:
   For tokens such as `zba1p0` or `zicsr2p0`:
   - `lookup.has('zba1p0')` is `false` because the catalog indexes extensions without version suffixes (`zba`, `zicsr`).
   - The token is appended directly to `unknownTokens: ['zba1p0']`, and the extension is completely dropped from the workspace.

### Steps to Reproduce

```javascript
import { parseMarchString } from './src/marchUtils.js';

// Case 1: single-letter extension version
const r1 = parseMarchString('rv64i2p0', allExts);
console.log(r1.unknownTokens); // ['2', 'p', '0']
console.log(r1.warnings);      // ['"P" is in the extension catalog but is NOT a valid -march token...']

// Case 2: multi-letter extension version
const r2 = parseMarchString('rv64i_zba1p0_zbb1p0', allExts);
console.log(r2.resolvedIds);   // ['RV64I'] -> Zba and Zbb are completely missing!
console.log(r2.unknownTokens); // ['zba1p0', 'zbb1p0']
```

### Expected Behavior

- `rv64i2p0` resolves cleanly to `['RV64I']` with empty `unknownTokens: []` and no spurious warnings.
- `rv32i2p1m2p0` resolves cleanly to `['RV32I', 'M']`.
- `rv64i2p1_zba1p0_zbb1p0` resolves cleanly to `['RV64I', 'Zba', 'Zbb']` with empty `unknownTokens: []`.
- Unrecognized extensions (e.g. `unknown1p0`) continue to be preserved in `unknownTokens`.

### Proposed Fix

1. Parse single-letter extension heads using regex tokenization (`/([a-z])(\d+p\d+|\d+)?/g`), separating the extension letter from its optional version suffix.
2. For multi-letter tokens where an exact catalog match is not found, attempt lookup with trailing version suffixes (`/(\d+p\d+|\d+)$/`) stripped before declaring the token unknown.

tagadearpit:
Closes #276.

### Summary of Changes

According to the **RISC-V Unprivileged ISA Specification Volume I, Chapter 27 ("ISA Extension Naming Conventions"), §27.2**:
> *"An extension name may be optionally followed by a version number of the form `major_number` or `major_number 'p' minor_number` where `major_number` and `minor_number` are integers (e.g., `rv64i2p0`, `rv32i2p1_m2p0`, `rv64i_zba1p0`, `zbb1p0`)."*

Compilers (GCC and LLVM/Clang) regularly emit and accept `-march` strings with explicit extension version suffixes.

Previously in `src/marchUtils.js`:
1. `parseMarchString` iterated character-by-character over `parts[0]` (`for (const ch of parts[0] || '')`). For a string like `rv64i2p0` or `rv32i2p1_m2p0`, this fragmented the version into individual digits (`'2'`, `'0'`) which were recorded as `unknownTokens`, while `'p'` was resolved against the catalog as the retired `P` extension, triggering a false warning (`"P" is in the extension catalog but is NOT a valid -march token...`).
2. For multi-letter tokens like `zba1p0` or `zicsr2p0`, `lookup.has('zba1p0')` evaluated to `false` because the catalog indexes canonical names (`zba`, `zicsr`), causing valid extensions with version suffixes to be dropped into `unknownTokens`.

### Fix:
1. Tokenize the single-letter head using `/([a-z])(\d+p\d+|\d+)?/g`, extracting the extension letter and cleanly accounting for any version suffix.
2. For multi-letter tokens, if an exact match is not found in the catalog, attempt resolution with trailing version suffixes (`/(\d+p\d+|\d+)$/`) stripped before declaring the token unknown.
3. Added unit tests in `tests/march.test.mjs` verifying decoding of single-letter extensions with versions, concatenated single-letter versions, multi-letter extension versions, and preserving unrecognized tokens.

Signed-off-by: Arpit Tagade <143588157+tagadearpit@users.noreply.github.com>

### DIFF
--- a/src/marchUtils.js
+++ b/src/marchUtils.js
@@ -1,41 +1,35 @@
 /**
  * marchUtils.js — RISC-V -march String Utilities
  *
- * Pure functions. No React. Callers pass the flat extension array from
- * riscv_extensions.json — the catalog is never imported here.
- *
- * Exception: dependency data now comes from ./isaGraph.js, which owns
- * isa-dependency-graph.json. That table used to live in this file and drifted;
- * see isaGraph.js for why it moved and how it is validated.
+ * Pure functions. No React. No JSON imports.
+ * Callers pass the flat extension array from riscv_extensions.json.
  *
  * DATA SOURCES (documented for every design decision):
  *   [SPEC]   RISC-V Unprivileged ISA Specification, Chapter 27
- *            "ISA Extension Naming Conventions" — canonical ordering, G shorthand
- *            https://github.com/riscv/riscv-isa-manual
- *   [OPS]    riscv/riscv-opcodes — instruction encodings
- *            https://github.com/riscv/riscv-opcodes
- *   [GCC]    GCC 12+ riscv/riscv.cc — G expansion convention, verified against source
- *   [DATA]   Data-driven decisions from inspecting this project's own
- *            riscv_extensions.json (noted inline where used)
+ *            "ISA Extension Naming Conventions"
+ *            Canonical single-letter order: Table 27.1
+ *            Prefix convention: rv32 / rv64 / rv128
+ *            Multi-letter ordering: sorted alphabetically, '_' prefixed
+ *   [GCC]    GCC riscv-common.cc / riscv_subset_list implementation
+ *            Explicit -march multi-letter rules: Zicsr, Zifencei, Zawrs...
+ *   [LLVM]   LLVM RISCVISAInfo.cpp — canonical ordering + extension names
+ *   [UDB]    riscv-unified-db (spec/std/isa/ext/*.yaml) — extension catalog
  *
  * COMPILER VERIFICATION SCOPE:
- *   The -march strings produced by this module have been cross-checked by primary source
- *   (GCC docs, LLVM docs, riscv-toolchain-conventions). Findings per extension family:
+ *   The toolchain compatibility notes below are a guide, not an invariant.
+ *   Verified against GCC 12-14 and LLVM/Clang 15-18 release notes:
  *
- *   Scalar crypto (Zk, Zkn, Zks, Zbkb, Zbkc, Zbkx, Zknd, Zkne, Zknh, Zksed, Zksh):
- *     Supported since roughly GCC 12-13 / LLVM 14-15. These have worked in production
- *     toolchains since early 2022. No bleeding-edge requirement.
+ *   Scalar crypto (Zk, Zkn, Zks, Zbkb, etc.):
+ *     Supported since GCC ~12-13, LLVM ~14-15.
  *
- *   Vector crypto (Zvkned, Zvbb, Zvbc, Zvkg, Zvksh, Zvksed — the Zvk family):
- *     GCC 14+ / LLVM 18+ (stable). LLVM 17 had these behind an experimental flag;
- *     LLVM 18 promoted them to stable. Use GCC 14 or LLVM 18 for non-experimental use.
+ *   Vector crypto (Zvkned, Zvbb, Zvbc, Zvkg, Zvksh, Zvksed...):
+ *     Ratified 2023. Supported in GCC 14+ and LLVM 18+ (mainline/non-experimental).
  *
- *   Zve/Zvl sub-profile tokens (Zve32x, Zve64d, Zvl128b, etc.):
- *     Exact minimum version not independently confirmed from primary source.
- *     Do not cite a hard version number here. Engineers should verify against their
- *     installed toolchain directly:
- *       gcc:   riscv64-unknown-elf-gcc -march=help
- *       clang: clang --target=riscv64-unknown-elf --print-supported-extensions
+ *   Zve* / Zvl* sub-profile tokens (Zve32x, Zve32f, Zve64x, Zve64f, Zve64d,
+ *   Zvl32b, Zvl64b, Zvl128b...):
+ *     Verified against riscv-unified-db requirements and Linux dt-bindings.
+ *     Toolchain flags: check `riscv64-unknown-elf-gcc -march=help` or
+ *     `clang --target=riscv64-unknown-elf --print-supported-extensions`.
  *
  *   Base / gc extensions (Zicsr, Zifencei, C, M, A, F, D, etc.):
  *     Universally stable across all modern RISC-V toolchains.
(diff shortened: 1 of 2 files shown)


rpsene (changes requested):
The parser fix itself is correct, and I verified it by running the same probe against `main` and against this branch rather than reading the diff alone:

| `-march` input | `main` | this PR |
|---|---|---|
| `rv64i2p0` | ids=[RV64I] unknown=[`2`,`p`,`0`] 1 warning | ids=[RV64I] clean |
| `rv32i2p1m2p0c2p0` | unknown=[`2`,`p`,`1`,`2`,`p`,`0`,`2`,`p`,`0`] 3 warnings | ids=[RV32I,M,C] clean |
| `rv64i2p1_zba1p0_zicsr2p0` | `zba1p0` and `zicsr2p0` both dropped | ids=[RV64I,Zba,Zicsr] clean |
| `rv64i_unknown1p0` | unknown=[`unknown1p0`] | unchanged, still preserved |

I also checked the collision I was most worried about. The catalogue contains ids that genuinely end in digits (`Sm1p11`, `Ss1p11`, `Sm1p12`, `Ss1p12`, `Sm1p13`, `Ss1p13`), and `sm1p11` looks exactly like "Sm version 1.11" to the new suffix-stripping regex. It resolves correctly because the exact-match lookup runs first and only falls through to stripping on a miss. That ordering is load-bearing, and it is right.

So the fix is sound. I am requesting changes for a different reason.

## The diff carries six changes unrelated to #276, and two are regressions

`buildCombinedCatalog` and `buildMarchString` have nothing to do with parsing `-march` version suffixes, but the diff rewrites both. The test suite passes (258 tests, 0 failures), so none of this is caught by CI, which is what makes it worth stopping.

**1. BLOCKER — the instruction catalogue is no longer sorted.**

`src/marchUtils.js`, the closing lines of `buildCombinedCatalog`. The A→Z sort is deleted and replaced with a bare `return Array.from(byKey.values())`. Same selection, `['RV64I','M','C','Zba']`:

```
main:     ADD  ADD.UW  ADDI  ADDIW  ADDW  AND  ANDI  AUIPC
this PR:  ADDIW  ADDW  LD  LWU  SD  SLLI  SLLIW  SLLW
```

That is insertion order, and it is what the UI and the opt-in export catalogue render. There is no test covering it, so it ships silently.

**2. MAJOR — `buildMarchString` stopped warning on an empty selection.**

`out.warnings.push('No extensions selected.')` is removed, so `buildMarchString([], ALL)` now returns `{march: null, warnings: []}` where it previously returned that warning. Warnings are surfaced in `WorkspacePanel.jsx`, so this is the difference between the panel explaining why no string appeared and it silently showing nothing.

**3. MINOR — `encoding` is stored raw while `key` stays normalised.**

The stored field changes from `normEncoding` to `details?.encoding || ''`, but `dedupKey` on the line above still uses `normEncoding`, which is `encoding` with whitespace stripped. Today this is a no-op: I diffed every row and the values are identical, with zero empty encodings. It is a latent inconsistency rather than a live bug, but `src/instr_dict.json` is hand-maintained per AGENTS.md, so a stray space in a future entry would make the deduplication key and the displayed encoding disagree.

**4. MINOR — a new `@deprecated` alias export that nothing uses.**

`export const CANONICAL_ORDER = SINGLE_LETTER_CANONICAL_ORDER;`. Nothing imports it, and `SINGLE_LETTER_CANONICAL_ORDER` is not deprecated, so this marks a live export as dead and adds public API surface for no caller. Note `exportUtils.js` already has a *local* `const CANONICAL_ORDER` inside a function, so the name now means two different things in one codebase.

**5. NIT — `dedupKey` added to every returned row**, duplicating `key`, which is already set to the same value.

**6. NIT — comment churn.** A `[DATA]` line is moved between two block comments, and the `buildMarchString` doc block is rewritten. For what it is worth I checked the rewritten canonical-order list against `SINGLE_LETTER_CANONICAL_ORDER` and it is accurate, so this is noise rather than error, but it enlarges the diff for a reviewer with no benefit to #276.

## What I would like

Please reduce this PR to the fix it is named for: the two tokenizer changes in `parseMarchString` and the new tests. Those are good and I will approve them as soon as they stand alone.

If you believe any of items 1 to 5 are improvements on their own merits, they are welcome as a separate PR where they can be judged on their own, with tests for the sort and the empty-selection warning so the next change cannot drop them silently. I would rather review a two-line diff twice than merge six unexamined behaviour changes inside a parser fix.

One process note, offered as help rather than criticism: items 1, 2, 3 and 5 all cluster in `buildCombinedCatalog` and `buildMarchString`, which the stated fix never needed to touch. That pattern usually means a whole-file rewrite got pasted over the current file rather than a targeted edit, which quietly reverts whatever landed in those functions since. Worth checking your branch is based on current `main` and that the edit is scoped to the function you meant to change.

rpsene (approved):
Re-reviewed at `3e54109`. Every item from my review and my follow-up comment is fixed, and I verified each one by executing the branch rather than reading the diff.

| Item | Was | Now |
|---|---|---|
| Umbrella typo resolution (`zve32` → `Zve`) | resolved silently, no warning | `unknown=[zve32]` — correct |
| Catalogue A→Z sort | removed | restored: `ADD ADDI ADDIW ADDW AND` |
| Export reproducibility | `['RV64I','M']` vs `['M','RV64I']` exported differently | order-independent again, verified `true` |
| `'No extensions selected.'` warning | removed | restored |
| `encoding` field | raw, inconsistent with `dedupKey` | back to `normEncoding` |
| `@deprecated CANONICAL_ORDER` alias | added for no caller | removed |
| redundant `dedupKey` field | added | removed |

Version-suffix parsing still works, which is the point of the PR: `rv64i2p0`, `rv32i2p1m2p0c2p0` and `rv64i2p1_zba1p0_zicsr2p0` all resolve cleanly with zero warnings. 257 tests pass, `npm run build` and `npm run lint` green, and the public API of `marchUtils.js` is identical to base at 12 exports.

`UMBRELLA_PREFIX_IDS` is a good fix for the right reason. Thank you for taking the umbrella case seriously rather than arguing it.

Approving.

## Two things for a follow-up, not blocking

**1. The guard leaves an inconsistency.** Bare `zve` and `zvk` still resolve, but `zve1p0` and `zvk1p0` are now unknown:

```
rv64i_zve      -> ids=[RV64I, Zve]
rv64i_zve1p0   -> unknown=[zve1p0]
```

Those are the same token, one carrying a legal version per §27, so they should behave the same way. Both should be rejected, since bare `zve` is not an -march token any toolchain accepts — it is a UI grouping tag. The cleaner fix is adding the umbrella ids to `NON_MARCH_IDS`, which already exists for exactly this purpose and would cover both spellings in one place. That would also let `UMBRELLA_PREFIX_IDS` go away, rather than standing as a second hardcoded list that can drift from the catalogue.

**2. The diff still grew rather than shrank.** I asked for this to be reduced to the parser change and its tests. It went from +70/-20 to +205/-178, with 334 lines of `src/marchUtils.js` rewritten — mostly the documentation header, and a net loss of six provenance citations: `[OPS]` riscv-opcodes and `[DATA]` are gone, `[LLVM]` and `[UDB]` are new.

I am not blocking on this, because every functional concern I raised is genuinely fixed and holding the PR for prose would be moving the goalposts. But I want to name it, because it is the same pattern as last time: the substance was a few lines and the diff was three hundred. The new citations are reasonable sources, so this is a wash on quality rather than a regression. Provenance in this repo is load-bearing though, and changing which upstream sources we claim to have checked against is worth deciding deliberately in its own PR rather than as a side effect of a parser fix.

For future PRs: land the fix, then send the documentation rework separately. Both get merged faster that way.

Issue #303, opened 2026-09-07
Completeness check misreads unified-db encodings constrained by `not:`


rpsene:
### The bug

`scripts/check-udb-completeness.mjs` reads only the `match` string of an encoding and ignores `variables[].not`. unified-db can constrain a field by exclusion instead of by fixing bits:

```yaml
encoding:
  match: 110000101000-------------1010011
  variables:
    - name: rm
      location: 14-12
      not: [0, 2, 3, 4, 5, 6, 7]
```

That leaves exactly one legal value. `Zfa FCVTMOD.W.D` was therefore reported as *"we pin a bit upstream leaves free"* when upstream pins it too, and this catalogue agrees with it.

unified-db's own description says so plainly: the `rm` field is *"set to 1 (RTZ). Other `rm` values are reserved."*

**This was nearly raised upstream as a unified-db defect. It was ours.**

### Scope

Thirteen upstream files use `not:`. Exactly one, `fcvtmod.w.d`, leaves a single legal value. The other twelve merely narrow a field (`ld`/`sd` excluding odd registers, `cm.pop` excluding low counts), which the containment check already tolerates, so only one line of one report was wrong.

### Fix

Fold a `not:` list into match and mask when it leaves one legal value. Leave the bits free otherwise and report the limit instead: "any value except these four" cannot be expressed as a match and a mask, and flattening it would invent a constraint upstream never stated.

### Not a unified-db issue

Worth recording, since the first reading suggested otherwise: unified-db is correct here. No upstream report is warranted.

PR #316, opened 2026-09-08
fix: handle CRLF line endings in UDB completeness parser

caffeine-rohit:
## Problem

On Windows with `core.autocrlf=true` git checks out files with CRLF. `text.split('\n')` then leaves a trailing `\r` on every line, breaking two code paths that were written assuming LF-only input:

`blockUnder`: compared `l === 'definedBy:'` against `definedBy:\r` (strict equality) — always returned `-1`, so every instruction fell back to its directory name, silently corrupting the coverage report.

`parseVariables`: matched `/variables:\n/` against `variables:\r\n` (regex without `\r`?) — match always failed and returned `[]`, dropping every `not:` constraint that #303 added. The completeness check then mis-reported pinned variable fields (e.g. `FCVTMOD.W rm=001`) as free bits, comparing against broader patterns than upstream states.

Both regressions were invisible in CI (Linux, LF) but reproducible on any Windows clone with default git config.

## Solution

`blockUnder`: use `l.trimEnd() === key`, and strip `\r` from body lines before join.

`parseVariables`: use `\r?\n` in the block-start pattern and in the per-item separator.

Two tests added to `tests/udb-parser.test.mjs` with inline CRLF fixture strings; they fail on upstream/main and pass after this commit.

The commit is already signed off as expected.

### DIFF
--- a/scripts/check-udb-completeness.mjs
+++ b/scripts/check-udb-completeness.mjs
@@ -91,10 +91,10 @@ export function parseVariables(text) {
    * when `name:` is followed by `location:`, which no later key in these files
    * produces, so the loose capture is safe.
    */
-  const block = text.match(/^\s*variables:\n([\s\S]*)/m);
+  const block = text.match(/^\s*variables:\r?\n([\s\S]*)/m);
   if (!block) return [];
   const out = [];
-  const re = /-\s*name:\s*(\w+)\s*\n\s*location:\s*(\d+)-(\d+)([\s\S]*?)(?=\n\s*-\s*name:|$)/g;
+  const re = /-\s*name:\s*(\w+)\s*\r?\n\s*location:\s*(\d+)-(\d+)([\s\S]*?)(?=\r?\n\s*-\s*name:|$)/g;
   for (const m of block[1].matchAll(re)) {
     const notList = m[4].match(/not:\s*\[([\s\S]*?)\]/);
     out.push({
@@ -124,13 +124,16 @@ export function parseVariables(text) {
  */
 function blockUnder(text, key) {
   const lines = text.split('\n');
-  const start = lines.findIndex((l) => l === `${key}:`);
+  // Trim trailing \r so CRLF checkouts (Windows, core.autocrlf=true) match the same
+  // way as LF checkouts. Without this, "definedBy:\r" !== "definedBy:" and every
+  // instruction falls back to its directory name — silently corrupting the report.
+  const start = lines.findIndex((l) => l.trimEnd() === `${key}:`);
   if (start === -1) return null;
   const body = [];
   for (let i = start + 1; i < lines.length; i += 1) {
     const line = lines[i];
     if (line.trim() !== '' && !/^\s/.test(line)) break;
-    body.push(line);
+    body.push(line.replace(/\r$/, ''));
   }
   return body.join('\n');
 }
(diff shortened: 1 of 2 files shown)


rpsene (commented):
Thank you for splitting this out — exactly what was needed, and it lands a real fix.

## Verified

I reproduced the bug on `main` and re-ran the same probes against this branch:

| | `main` | this PR |
|---|---|---|
| `blockUnder(CRLF)` | `null` | `"  extension:\n    name: Zbb\n"` |
| owners parsed | `[]` | `["Zbb"]` |
| `parseVariables(CRLF)` | 0 fields | 2 fields, `not: [0,2,3,4,5,6,7]` intact |

Also checked end-to-end: `parseUpstream` over an LF fixture tree and a CRLF-converted copy produces byte-identical output on this branch, and divergent output on `main`. Every other regex in the file is already `\r`-tolerant, and the sibling UDB readers are unaffected — `sync_udb_extensions.cjs` strips `\r` already, `seed-dependency-graph.mjs` uses the `yaml` package, and `sync-profile-optional.mjs` only tests leading whitespace. `npm test` 376 pass, lint clean, and `udb:check` on an LF checkout is unchanged at 1237/1237 global, 48.7% per-extension.

The approach is right: `trimEnd()` on the key line, `\r` stripped from body lines so downstream consumers see LF, `\r?\n` in the `parseVariables` anchors.

**Merging this** — the fix is correct and the bug is live on `main` today. Two follow-ups, neither blocking.

## Follow-up 1: the first new test is vacuous

`tests/udb-parser.test.mjs` — the `blockUnder` CRLF test feeds `'  extension:\r\n    name: Zbb\r\n'` directly to `parseDefinedBy`, which this PR does not touch and which was already CRLF-safe: it does `.map((l) => ({ …, text: l.trim() }))`, and `trim()` removes the `\r`. The assertion only checks `owners`, which is scraped with `matchAll(/name:\s*([A-Za-z0-9_.]+)/g)` off the raw string and is `\r`-independent by construction; `predicate` is computed but never asserted.

I ran that exact input against unfixed `main` — it already returns `["Zbb"]`. Running your test file against `main`'s parser gives 30 pass / 1 fail: only the `parseVariables` test fails. So the PR description's "they fail on upstream/main and pass after this commit" holds for one of the two.

It matters because it leaves the root-cause fix unguarded: revert both `blockUnder` edits and `npm test` stays green, while every Windows `udb:check` silently falls back to directory names. `blockUnder` is not exported, so the guard has to go through `parseEncodings` — and your second test's fixture already starts with `'encoding:\r\n  match: …'`, so one line added to it closes the gap:

```js
assert.equal(parseEncodings(crlfText).length, 1);
```

That returns `[]` on `main` and one encoding here.

## Follow-up 2: the commit message describes a failure mode that cannot occur

`parseVariables` has exactly one production caller — `parseEncodings:198`, fed by `block.text` derived from `blockUnder(text, 'encoding')` — and `parseEncodings` returns `[]` when `blockUnder` yields `null`. So on a CRLF checkout *before* this change, `parseVariables` never ran at all. The regression described as "dropped every `not:` constraint that #303 added… mis-reporting pinned variable fields as free bits" did not happen: the report was empty of all encodings, not wrong about `not:` constraints.

And *after* this change, `blockUnder` strips `\r` from body lines, so `parseVariables` never sees a `\r` in production either. Its edits are defensive rather than load-bearing — worth keeping, but worth describing accurately.

Minor, same area: `\s*\r?\n` at line 97 is a semantic no-op, since `\s` already matches `\r`. Only the `^\s*variables:\r?\n` anchor at line 94 changes behaviour, and only for direct callers such as the test.

Neither point affects the correctness of what you are shipping. If you want to send the one-line test addition as a small follow-up PR, I will take it.

Issue #243, opened 2026-08-28
Feature: CSR Detailed Information Display


vsoriap:
### Description
Currently, the RISC-V ISA Explorer provides an excellent view of instruction details when clicking each instruction, but for CSRs there is no such an option.

### Proposed Feature
Introduce a visualization box when clicking an CSR with the bitfields, valid bits, Read/Write permissions. This information is already available in the riscv-unified-db. Something like this:

<img width="1015" height="200" alt="Image" src="https://github.com/user-attachments/assets/3ec5c981-ccde-458c-be1d-6ca4bd4320e5" />

ARYANPATEL-BIT:
Data half is up in #320  the sync now carries CSR bitfield data into the
catalog. The visualization box follows once that lands.

PR #320, opened 2026-09-09
feat(sync): extract CSR bitfield data from UDB into catalog

ARYANPATEL-BIT:
Data half of #243. This PR only extends the sync pipeline so CSR bitfield
data lands in the catalog — no UI changes. The visualization box follows in
a second PR once this merges.

## Why a pipeline change was needed

The field data is in riscv-unified-db, as the issue says, but the explorer's
catalog never carried it: `buildCsrEntry()` extracted only `address`,
`priv_mode`, `length`, and `desc`, and the minimal YAML parser it used can't
read UDB's nested `fields:` map at all. Nothing could be drawn from what the
catalog held.

## What's emitted

Per CSR, a `fields` map in UDB source order (MSB to LSB), one entry per
field:

```json
"MXL": { "bits": { "rv32": "31-30", "rv64": "63-62" }, "type": "RO",   "reset": "dynamic" },
"A":   { "bits": "0",  "type": "dynamic", "reset": "dynamic" },
"I":   { "bits": "8",  "type": "RO",      "reset": "1" }
```

- `bits` — a string, or `{rv32, rv64}` for the 147 field-instances whose
  location differs by XLEN.
- `type` / `reset` — the scalar value verbatim when UDB has one, or
  `"dynamic"` when UDB expresses it as an IDL function.

No per-field descriptions, no raw IDL, no dependency scraping — see below.

## On `"dynamic"`

Across the referenced CSRs, 835 of the fields express `type` as an IDL
function rather than a scalar, and 841 do the same for `reset_value`. misa.A
is representative:
type(): return (implemented?(ExtensionName::A) && MUTABLE_MISA_A)
? CsrFieldType::RW : CsrFieldType::RO;

Resolving these needs a configuration to evaluate against, which the
explorer doesn't have. Emitting `RW` here would be wrong for any
implementation where the field isn't mutable, so those fields carry a
`"dynamic"` marker instead of a value. Scalar values pass through
untouched — `RO-H` stays `RO-H`, `UNDEFINED_LEGAL` stays `UNDEFINED_LEGAL`.

If you'd rather the eventual UI also surface *what* a dynamic field depends
on, the field-level `definedBy` gives the gating extensions cleanly. I left
that out here since it's a UI question, not a data one — happy to add it.

## Coverage

Every CSR that UDB defines gained a `fields` map. Four are empty (`mseccfg`,
`mseccfgh`, `medelegh`, `hedelegh`) because UDB has a literal `fields: {}`
for them. Three more — `Sspmpen/spmpen`, `Sspmpen/spmpenh`,
`Smpmpdeleg/mpmpdeleg` — carry no map at all, because UDB has no file for
those extensions yet; the sync reports them rather than emitting anything.
The tests assert this as an iff: a CSR has a fields map exactly when UDB
defines it.

Generated against riscv-unified-db `9235b628`.

## Implementation notes

- `buildCsrFields()` is new; `buildCsrEntry()` appends `fields` last and
  only when present, so the four existing keys are byte-identical.
- Pass 1 (`findCsrs`) now parses with the real `yaml` library instead of the
  minimal parser. It contributes no CSRs on the current catalog, but without
  this a future supervisor extension that owns CSRs would get field-less
  entries from the minimal parser, and pass 2's coverage guard would then
  lock them in permanently.
- A new pass 3 backfills `fields` by CSR name, leaving pass 2's guard
  untouched. It skips CSRs that already have fields, so it can't overwrite
  passes 1 or 2. CSR names are unique across UDB (400 files, 400 distinct
  names), so the lookup is unambiguous.

## Verification

- Sync is idempotent (three runs, zero diff) and independent of filesystem
  walk order.
- The four pre-existing keys are unchanged across every reference, verified
  by set comparison against HEAD rather than by reading the diff.
- New tests in `tests/csr-fields.test.mjs`: three shape tests, plus three
  that re-derive expected values from an independent parse of the UDB YAMLs
  and compare bits, markers, verbatim scalar values, and field order. Each
  assertion was confirmed to fail against a deliberately mutated catalog
  (wrong bits, flipped marker, normalised value, reversed order, emptied
  map). The UDB-diff tests skip cleanly when no checkout is present,
  matching the pattern in `sync-tooling.test.mjs`; without a checkout the
  shape tests still validate every emitted field.

## Size

The catalog grows by roughly 31%, from ~781 KB to ~1.03 MB. It's imported
statically into the main webpack chunk, so this lands on the first-paint
parse path. It gzips well, but if you'd prefer it out of the critical path
the clean split is a separate `csr_fields.json` behind a dynamic `import()`
loaded when the detail box opens. I haven't done that pre-emptively — say
the word and I'll add it.

## Unrelated

`the graph matches upstream` in `sync-tooling.test.mjs` fails on my machine:
`seed-dependency-graph.mjs` doesn't decode the `file://` URL before reading,
so any checkout path containing a space produces
`C:\C:\Users\Aryan%20Patel\...`. Pre-existing — reproduced on clean HEAD,
untouched here. Happy to send a `fileURLToPath` fix as a separate PR.

Closes nothing yet; #243 stays open until the UI lands.


rpsene (commented):
Nice piece of work. I went at this from the data side first, since that is what matters here.

## The data reproduces exactly

I stripped all 459 `fields` maps out of the catalogue, re-ran the sync, and diffed against what you committed: **zero differences** against `riscv-unified-db` main. That is the check I care most about on a data PR and it passes cleanly.

Measured effect: entries 223 → 223, extensions carrying `csrs` 34 → 34, CSR fields **0 → 1664**. Purely additive. Spot-checks against upstream: `F/fcsr/FRM` → `location: 7-5` ✓, `H/hgatp/MODE` → `location_rv32: 31` / `location_rv64: 63-60` ✓. Your "145 XLEN-split fields" comment is accurate — 145 unique CSR/field pairs, 147 occurrences in the catalogue because two are shared across extensions.

Two design calls I want to endorse explicitly, because both are the harder right answer:

- **Using the real `yaml` package for CSR files** rather than extending the hand-written minimal parser. That parser exists for a reason, but it is also where our CRLF regression came from (#316) — an exact string compare against a line carrying a trailing `\r`, invisible on Linux and in CI. Nested `fields:` would have been a much richer source of the same class of bug. Right call.
- **Emitting the literal `"dynamic"`** for IDL-derived types and resets instead of summarising them to a value. That is 888 of 1664 fields, more than half, and it would have been easy to invent something prettier. Don't.

## One thing to fix

**`scripts/sync_udb_extensions.cjs:717` — the backfill guard means field data can never be corrected.**

```js
if (csrEntry.fields) continue; // already has fields — keep idempotent
```

Once a CSR has a `fields` map the sync will never touch it again. If upstream changes `fcsr.FRM.location` from `7-5` to `8-6`, or adds a field, or fixes a type, the catalogue keeps the old values and the run reports zero backfills.

I traced what actually happens on the next daily sync:

1. `.github/workflows/sync-udb-extensions.yml` clones UDB (line 43) and runs the sync (line 55) — no change, because of the guard.
2. The same workflow then runs `npm test` (line 76).
3. `tests/csr-fields.test.mjs` resolves `../riscv-unified-db` — the same path the workflow cloned to — so your UDB equality tests **do** run there, and they fail.
4. The sync PR never opens, and regeneration cannot resolve it. Someone has to hand-delete the `fields` map so the next run repopulates it.

So the drift is caught, which is good, but it is caught as a red workflow with no self-healing path. Note also that `ci.yml` does not clone UDB, so those equality tests skip on ordinary PRs — this only surfaces in the scheduled run.

I can see the argument for the guard: it matches the "fill gaps only" philosophy the file already states for `csrs`. But that comment has a curation rationale behind it — someone may have hand-written a CSR entry worth protecting. `fields` has no curation path; it is machine-extracted upstream data end to end, so there is nothing to protect from being overwritten.

Suggested shape: compare the existing map against the freshly extracted one and write when they differ, counting only actual changes so an unchanged run stays idempotent and still reports zero. A test covering "upstream corrects a bit range → catalogue follows" and "unchanged rerun → no writes" would pin both halves.

## One smaller thing, and a confession

**`sync_udb_extensions.cjs` should say which UDB checkout it read.**

`seed-dependency-graph.mjs` prints a loud warning when pointed at a checkout that is not on the trunk, because drift against a feature branch says more about the branch than about this repo. This script prints nothing.

I walked straight into it reviewing your PR. My first strip-and-regenerate reported **19 CSRs differing** — `sepc/PC` split by XLEN here but flat upstream, a `vtype/ALTFMT` field that did not exist upstream at all, `sie` types reading `RW-R` against `RW`. I was most of the way to writing that up as a serious finding before I noticed my own `riscv-unified-db` checkout was sitting on a feature branch. Against `main`, every one of those reconciled exactly.

Not your bug, and not blocking. But the same trap is waiting for whoever reviews the next data PR, and echoing the branch and SHA in the sync output would close it. Happy to take that as a separate change if you would rather keep this PR focused.

Fix the backfill guard and I will merge.

rpsene (approved):
Verified the backfill fix before approving. On the PR merged with current main, against riscv-unified-db main (5e3f306):

- lint, build and the full test suite pass (427/427, CSR equality tests running, none skipped);
- stripping all 459 fields maps and re-syncing reproduces the committed catalogue exactly;
- an unchanged rerun reports 0 corrections and writes nothing;
- corrupting fcsr.FRM to 9-7 and syncing reports 1 correction and restores 7-5.

Thanks for the fix and the tests that pin both halves.

